PR #26, opened 2016-06-01
Update AppDelegate.swift

diamo1213:
FIRInstanceID.instanceID().token()! cause crash the app from the quick-start example when the app is installed and launched first time on a real device.


kroikie:
@diamo1213 I'd love to accept your PR, could you sign the Contributor license, then I'd be happy to merge it.

kroikie:
@diamo1213 also, looks like you don't need to define the refreshedToken as a string, so just removing the ! should get the desired result.

PR #30, opened 2016-06-02
Updated selector to new swift syntax. Fixed typo. Fixed crash when FI…

rcpfuchs:
…RInstanceID.instanceID().token() was force unwrapped, after connection fail.

Line 43. Changed tokenRefreshNotificaiton to tokenRefreshNotification and used #selector as suggested by Xcode. With previous syntax I did not receive the notification.

Line 65 - 70. Unwrapped refreshedToken with if let, because the force unwrap caused a crash for me when the connection to FCM was not successful and the token was nil.

### DIFF
--- a/messaging/FCMSwift/AppDelegate.swift
+++ b/messaging/FCMSwift/AppDelegate.swift
@@ -40,7 +40,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     FIRApp.configure()
 
     // Add observer for InstanceID token refresh callback.
-    NSNotificationCenter.defaultCenter().addObserver(self, selector: #selector(self.tokenRefreshNotificaiton),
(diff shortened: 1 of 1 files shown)


diamo1213:
Hi @rcpfuchs,

In line 43: Type 'AppDelegate' has no member 'tokenRefreshNotification'

rcpfuchs:
Hey @diamo1213 is that an error you get? Because 

func tokenRefreshNotification(notification: NSNotification) {
      if let refreshedToken = FIRInstanceID.instanceID().token() {
          print("InstanceID token: \(refreshedToken)")
      }
      
      // Connect to FCM since connection may have failed when attempted before having a token.

... Is the member, or not?

diamo1213:
@rcpfuchs ok, your decision works too now

iulukaya:
Address this on https://github.com/firebase/quickstart-ios/pull/73/commits/321fdf1fab515628eb06d084777c57843bbfbf46

Kept the selector as typo was fixed before.

PR #52, opened 2016-06-24
Replace sharedApplication to method's argument

mbelsky:
Hi,
This method application:didFinishLaunchingWithOptions provides an application instance, so sharedApplication is excess there


thetruthisfree777:
Fuck yes

morganchen12:
Gonna close this, since it seems too stale at this point to be worth salvaging. 

There are still two calls to `[UIApplication sharedApplication]` in `application:didFinishLaunchingWithOptions:`, but those are part of a code snippet that's also displayed elsewhere in isolation, so I'm ok with keeping them.

PR #84, opened 2016-08-31
Update README.md

a-b:
Add a list of subprojects


ianbarber:
Thanks! I like this idea, but I had a bunch of nits about the actual text, so I've gone ahead and made a separate set of changes. Great idea though!

Issue #103, opened 2016-09-20
No notifications in background on iOS 10 (Swift)


pimnijman:
I've implemented everything according to the guides and this example: https://github.com/firebase/quickstart-ios/blob/master/messaging/FCMSwift/AppDelegate.swift

I do receive messages when my app is in the foreground, but I don't when it's in the background.

When I remove the iOS 10 specific code like this …
   ```
    //if #available(iOS 10.0, *) {
    //  let authOptions : UNAuthorizationOptions = [.Alert, .Badge, .Sound]
    //  UNUserNotificationCenter.currentNotificationCenter().requestAuthorizationWithOptions(
    //    authOptions,
    //    completionHandler: {_,_ in })
    //
    //  // For iOS 10 display notification (sent via APNS)
    //  UNUserNotificationCenter.currentNotificationCenter().delegate = self
    //  // For iOS 10 data message (sent via FCM)
    //  FIRMessaging.messaging().remoteMessageDelegate = self
    //
    //} else {
      let settings: UIUserNotificationSettings =
      UIUserNotificationSettings(forTypes: [.Alert, .Badge, .Sound], categories: nil)
      application.registerUserNotificationSettings(settings)
      application.registerForRemoteNotifications()
    //}
```
… everything works as expected.
Please advice. All help is greatly appreciated.

lethianqt94:
I don't receive any messages in both foreground or background. :(

roughike:
https://github.com/firebase/quickstart-ios/pull/101

I struggled with this for days, then compared to the Objective-C sample that worked properly. The iOS 10 version is missing the ```registerForRemoteNotifications()``` function call. The pull request above should fix it.

pimnijman:
Thank you @roughike, you've saved me at least three gray hairs.
Is there any way I can upvote your pull request?

roughike:
@pimnijman Glad I helped!

Maybe add a 👍 reaction to the PR? :)

sir-gon:
I followed [this guide](http://stackoverflow.com/a/39632846/6366150) and it worked.

roughike:
@sir-gon Yeah, with Objective-C there's no problem. Try the Swift sample and you know what we're talking about.

@pimnijman Maybe change the title to say this happens with the Swift sample to avoid confusion and be more specific?

evanjmg:
[...]

Issue #102, opened 2016-09-20
Only foreground 


dimohamdy:
the notification only work when app in foreground not work in background or when app killed
i run in iOS 10 and Xcode 8

i use console to send message

dimohamdy:
i found the solution here
https://github.com/firebase/quickstart-ios/pull/101

PR #108, opened 2016-09-26
Updated to Swift 3 and fixes

JEKos:
Updated to Swift 3 and application.registerForRemoteNotifications() add to iOS 10 too

### DIFF
--- a/messaging/FCMSwift/AppDelegate.swift
+++ b/messaging/FCMSwift/AppDelegate.swift
@@ -16,7 +16,6 @@
 
 import UIKit
 import UserNotifications
-
 import Firebase
 import FirebaseInstanceID
 import FirebaseMessaging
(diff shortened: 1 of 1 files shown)


JEKos:
I signed it!

mcdonamp:
Thanks @JEKos, especially given how painful some migrations to Swift 3 have been! I don't think we've settled whether we're going to use Swift 3/XCode 8 by default for all new code, so we'll wait until we've synced internally to merge this. It might be fun to fork this into a Swift 3.0 scheme and keep it in parallel to the Swift 2.x one for devs on both (though TBH we're really not sure how much 2.x code exists in the world vs 3.x--assume it's a fair margin of existing development).

JEKos:
ok, but the `application.registerForRemoteNotifications()` for iOS 10 is important, notification doesn't work in iOS 10 without this line

morganchen12:
cc @iulukaya

roughike:
@mcdonamp What @JEKos said. The Swift sample is broken for iOS 10, as notifications don't come through when the app is in background. This was a major headache for our team.

[Here's a PR that fixes that, and that only](https://github.com/firebase/quickstart-ios/pull/101)

See also:
* https://github.com/firebase/quickstart-ios/issues/102
* https://github.com/firebase/quickstart-ios/issues/103

mcdonamp:
Cool, merged #101 (which now caused conflicts here), so hopefully people are unblocked on iOS 10 while we chat about Swift 3. Given that Swift 3 means XCode 8, and thus means most people will be using iOS 10, it's probably reasonable that we make this change, I think we're just trying to make sure we can update some of our other material (like our docs) to Swift 3 so it's consistent.

messaging/FCMSwift/AppDelegate.swift
```diff
@@ -16,124 +16,127 @@
 
 import UIKit
```
qkies: UIKit

messaging/FCMSwift/AppDelegate.swift
```diff
@@ -16,124 +16,127 @@
...
 import UserNotifications
-
 import Firebase
 import FirebaseInstanceID
```
qkies: UIKit

messaging/FCMSwift/AppDelegate.swift
```diff
@@ -16,124 +16,127 @@
 
 import UIKit
 import UserNotifications
-
 import Firebase
```
qkies: UIKit

iulukaya:
I think we addressed the remaining part in our swift 3 upgrade. Thanks for your contribution.

PR #115, opened 2016-09-30
Attempt to make Swift sign in code easier to read

peterkos:
When I was first learning the ins and outs of Firebase, I came here to get a real-world, full example of how the Firebase code worked.

However, I was appalled at what I saw. 13 levels of indentation? **13**!?
And most were just nested `if let` statements -- not a `guard` anywhere to be found.

This pull request is my attempt at turning at least one of the classes into a more readable example, with added comments to explain what is going on (hopefully more helpful to a beginner) and with some pieces of code broken up into bite-sized functions. 

Note: the spinner code probably isn't implemented properly. I can't figure out why you'd call a hideSpinner *closure* inside the startSpinner closure. In particular, a [certain progress indicator framework](https://github.com/SVProgressHUD/SVProgressHUD) uses functions to start and stop the spinner. Is there a technical reason why that procedure couldn't be implemented here, or is just a design decision?

---

Aside from those changes, I have a question about the overall design of the app.
Why does the "Create account" procedure use dialog popups and not a simple static TableViewController? As a user, it's hard to follow, with constant animations popping up and going with no distinction between them and errors, and the code contains closure after closure after closure.

### DIFF
--- a/database/DatabaseExampleSwift/SignInViewController.swift
+++ b/database/DatabaseExampleSwift/SignInViewController.swift
@@ -96,58 +96,94 @@ class SignInViewController: UIViewController, UITextFieldDelegate {
     
   }
   
-  
   @IBAction func didTapSignUp(_ sender: AnyObject) {
-    showTextInputPrompt(withMessage: "Email:") { (userPressedOK, email) in
-      if let email = email {
-        self.showTextInputPrompt(withMessage: "Password:") { (userPressedOK, password) in
-          if let password = password {
-            self.showTextInputPrompt(withMessage: "Username:") { (userPressedOK, username) in
-              if let username = username {
-                self.showSpinner({
-                  FIRAuth.auth()?.createUser(withEmail: email, password: password) { (user, error) in
-                    self.hideSpinner({
-                      if let error = error {
-                        self.showMessagePrompt(error.localizedDescription)
-                        return
-                      }
-                      self.showSpinner({
-                        let changeRequest = FIRAuth.auth()?.currentUser?.profileChangeRequest()
-                        changeRequest?.displayName = username
-                        changeRequest?.commitChanges() { (error) in
-                          self.hideSpinner({
-                            if let error = error {
-                              self.showMessagePrompt(error.localizedDescription)
-                              return
-                            }
-                            // [START basic_write]
-                            self.ref.child("users").child(user!.uid).setValue(["username": username])
-                            // [END basic_write]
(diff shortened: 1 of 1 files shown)


database/DatabaseExampleSwift/SignInViewController.swift
```diff
@@ -36,97 +36,150 @@ class SignInViewController: UIViewController, UITextFieldDelegate {
...
   @IBAction func didTapEmailLogin(_ sender: AnyObject) {
-    if let email = self.emailField.text, let password = self.passwordField.text {
-      showSpinner({
-        FIRAuth.auth()?.signIn(withEmail: email, password: password) { (user, error) in
-          self.hideSpinner({
-            if let error = error {
-              self.showMessagePrompt(error.localizedDescription)
+  
+    guard let email = self.emailField.text, let password = self.passwordField.text else {
+          self.showMessagePrompt("email/password can't be empty")
```
morganchen12: remove extra whitespace before this line

database/DatabaseExampleSwift/SignInViewController.swift
```diff
@@ -36,97 +36,150 @@ class SignInViewController: UIViewController, UITextFieldDelegate {
...
+          
+          self.showSpinner{}
+          
+          let changeRequest = FIRAuth.auth()?.currentUser?.profileChangeRequest()
+          changeRequest?.displayName = username
+          
+          changeRequest?.commitChanges() { (error) in
+            
+            self.hideSpinner{}
+            guard error == nil else {
```
morganchen12: If you're not binding any variables in this `guard` statement, why not just use an `if let`? That way you can avoid the (inconsequential, but ugly) force-unwrap later.

database/DatabaseExampleSwift/SignInViewController.swift
```diff
@@ -36,97 +36,150 @@ class SignInViewController: UIViewController, UITextFieldDelegate {
...
-          self.hideSpinner({
-            if let error = error {
-              self.showMessagePrompt(error.localizedDescription)
+  
+    guard let email = self.emailField.text, let password = self.passwordField.text else {
+          self.showMessagePrompt("email/password can't be empty")
+      return
+    }
+    
+    showSpinner{}
```
morganchen12: Be consistent with `showSpinner` and `self.showSpinner`. Also, please add a space between the method name and the curly braces so it's consistent with the rest of the file.

```swift
self.showSpinner {}
```

PR #176, opened 2016-11-29
Add check for message ID

kroikie:
Avoid crashes if non FCM messages, those without a message ID, are received.

### DIFF
--- a/messaging/MessagingExampleSwift/AppDelegate.swift
+++ b/messaging/MessagingExampleSwift/AppDelegate.swift
@@ -25,6 +25,7 @@ import FirebaseMessaging
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
   var window: UIWindow?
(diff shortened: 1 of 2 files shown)


ianbarber:
@iulukaya - can you verify these test fails are just flakiness? They don't look related to me, in which case we should merge.

messaging/MessagingExampleSwift/AppDelegate.swift
```diff
@@ -69,7 +70,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     // TODO: Handle data of notification
 
     // Print message ID.
-    print("Message ID: \(userInfo["gcm.message_id"]!)")
+    if userInfo[gcmMessageIDKey] != nil {
```
iulukaya: can you use "let" instead of !=nil
if let  gcmMessage = userInfo[gcmMessageIDKey] {
  print gcmMessage

PR #182, opened 2016-12-12
Fix code style

revolter:
The delegates are properties, not methods, so they must be accessed
using dot notation.


revolter:
I signed it!

morganchen12:
Thanks!

revolter:
This needs to be updated in the documentation too.

PR #186, opened 2016-12-13
Fix incorrect FCM connection logic

revolter:
Prior to this, when running the Example app for the first time, the
following error was being printed in the console:

Unable to connect to FCM. Error Domain=com.google.fcm Code=501 "(null)"

After accepting the notifications permission, the following error was
being printed in the console:

Unable to connect to FCM. Error Domain=com.google.fcm Code=2001 "(null)"

The first error means that the app tried to connect to FCM while there
was no Firebase registration token available. It now checks for the
current token and exits early if it's null.

The second error means that the app tried to connect to FCM while
another connection is already available. It now disconnects from any
existing connection prior to trying to connect again.

The better solution would be to have an `isConnected` property on the
`FIRMessaging` shared instance and if it is set to `YES`, exit early
instead of re-connecting to FCM.

This could also be achieved by a similar private property in the
`AppDelegate` class but I though that would be outside the example's
scope and having only parts of the code in the documentation would make
users not to understand it properly.

I also don't understand the reasoning behind these cryptic errors.
Firstly, the error codes should be added to an errors enum. Secondly,
an error message would immensely help, instead of those null values.

Even with these added, I don't understand the "call methods and ignore
the errors in case of problems" reasoning behind this Example, when they
could be avoided. It took me days to figure out what those errors really
mean.

### DIFF
--- a/messaging/MessagingExample/AppDelegate.m
+++ b/messaging/MessagingExample/AppDelegate.m
@@ -187,6 +187,13 @@ didReceiveNotificationResponse:(UNNotificationResponse *)response
 
 // [START connect_to_fcm]
 - (void)connectToFcm {
+  if (![[FIRInstanceID instanceID] token]) {
+    return;
+  }
+
+  // Disconnect previous FCM connection if it exists.
+  [[FIRMessaging messaging] disconnect];
+
   [[FIRMessaging messaging] connectWithCompletion:^(NSError * _Nullable error) {
     if (error != nil) {
       NSLog(@"Unable to connect to FCM. %@", error);
--- a/messaging/MessagingExample/AppDelegate.m
+++ b/messaging/MessagingExample/AppDelegate.m
@@ -187,6 +187,7 @@ didReceiveNotificationResponse:(UNNotificationResponse *)response
 
 // [START connect_to_fcm]
 - (void)connectToFcm {
+  // Won't connect since there is no token
   if (![[FIRInstanceID instanceID] token]) {
     return;
   }
--- a/messaging/MessagingExampleSwift/AppDelegate.swift
+++ b/messaging/MessagingExampleSwift/AppDelegate.swift
@@ -107,6 +107,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
   // [START connect_to_fcm]
   func connectToFcm() {
+    // Won't connect since there is no token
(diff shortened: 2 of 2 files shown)


kroikie:
Thanks for the PR, just a small change requested and them I'm happy to merge.

revolter:
What's the change?

revolter:
Also, this could also be handled by the `connectWithCompletion:` method internally.

kroikie:
Indeed the changes in this PR are handled by the `connectWithCompletion` method, while checking that the token is not nil before connecting will make the logs a bit cleaner it's not necessary.

So I'll close this PR and let `connectWithCompletion` handle things for this sample. Thanks for the PR though.

revolter:
@kroikie, No, they are not! I said that it will be nice if it did handle them.

With the current state of the SDK and the Example app, it is throwing errors in the console, which it shouldn't.
And also they're cryptic and misleading for newcomers, so they waste a lot of time figuring what's up.

If you don't want to merge this, then add them in the documentation (with the risk of looking bad), because people need to know about them.

messaging/MessagingExample/AppDelegate.m
```diff
@@ -187,6 +187,13 @@ - (void)tokenRefreshNotification:(NSNotification *)notification {
 
 // [START connect_to_fcm]
 - (void)connectToFcm {
+  if (![[FIRInstanceID instanceID] token]) {
+    return;
```
kroikie: Could you add a log message here that states "Won't connect since there is no token"?

messaging/MessagingExampleSwift/AppDelegate.swift
```diff
@@ -107,6 +107,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
   // [START connect_to_fcm]
   func connectToFcm() {
+    guard FIRInstanceID.instanceID().token() != nil else {
+      return;
```
kroikie: Same as Obj C comment.

kroikie:
@revolter I was wrong here, sorry my mistake. I was thinking of another method. This should be handled by the SDK, I filed an internal issue to update this, your PR can be used till that is fixed.

If you can add the log message that I requested then I'll be happy to merge.

revolter:
Done

PR #188, opened 2016-12-13
Add note about notification permission dialog

### LINKED ISSUE #184: Notifications sample show bad practice
I can't understand why is the Internet __FULL__ of bad practice examples about notifications registration. Moreso, why is even Firebase promoting this?

As you can see in [the sample's `AppDelegate.m`][1] and in [this documentation page][2], the registration is done at the app start. Is anyone working on real world applications? You don't request notifications permission at app start as no one will ever accept them, ever!

The documentation briefly states that you "can" request it at a desired point but the sample doesn't have at least any warning. Because of this, the `[[FIRMessaging messaging] setRemoteMessageDelegate:self];` usage is unclear. Does it really have to be called when registering for notifications? I guess not.

But let's say I have an existing app (99% cases) and I want to use Firebase. I already am registering for notifications at, say, level up. I come across this example that looks like I need to set the delegate in my custom method that registers for notifications, but, in fact, I just have to set it after configuring the app.

Also, the sample uses `#if defined(__IPHONE_10_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0` instead of `if ([UNUserNotificationCenter class])`.

  [1]: https://github.com/firebase/quickstart-ios/blob/master/messaging/MessagingExample/AppDelegate.m#L47
  [2]: https://firebase.google.com/docs/notifications/ios/console-device#register_for_remote_notifications

kroikie:
This sample registers for remote notifications on first run,
if this is not desirable it should be moved to a more appropriate
place so the dialog is shown at a more appropriate time.

Resolves #184

### DIFF
--- a/messaging/MessagingExample/AppDelegate.m
+++ b/messaging/MessagingExample/AppDelegate.m
@@ -44,7 +44,8 @@ NSString *const kGCMMessageIDKey = @"gcm.message_id";
 - (BOOL)application:(UIApplication *)application
     didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
 
-  // Register for remote notifications
+  // Register for remote notifications. This shows a permission dialog on first run, to
+  // show the dialog at a more appropriate time move this registration accordingly.
   if (floor(NSFoundationVersionNumber) <= NSFoundationVersionNumber_iOS_7_1) {
     // iOS 7.1 or earlier. Disable the deprecation warnings.
     #pragma clang diagnostic push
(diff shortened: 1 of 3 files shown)


morganchen12:
LGTM

messaging/MessagingExample/AppDelegate.m
```diff
@@ -73,8 +74,7 @@ - (BOOL)application:(UIApplication *)application
           | UNAuthorizationOptionBadge;
       [[UNUserNotificationCenter currentNotificationCenter]
           requestAuthorizationWithOptions:authOptions
-          completionHandler:^(BOOL granted, NSError * _Nullable error) {
-          }
+            
        ];
 
       // For iOS 10 display notification (sent via APNS)
```
revolter: Shouldn't these lines have a note that says something like:

> These delegates setters must remain in the `willFinishLaunchingWithOptions:` or `didFinishLaunchingWithOptions:` methods, regardless of the notifications registration location.

?

revolter: @kroikie, Am I right with the question?

PR #189, opened 2016-12-13
Add doc include for connecting to FCM

kroikie:
- Add include for connecting to FCM when app becomes active.
- Add call to completion handler in didReceiveRemoteNotification.

### DIFF
--- a/messaging/MessagingExample/AppDelegate.m
+++ b/messaging/MessagingExample/AppDelegate.m
@@ -125,6 +125,10 @@ NSString *const kGCMMessageIDKey = @"gcm.message_id";
 
   // Print full message.
   NSLog(@"%@", userInfo);
+
+  if (completionHandler) {
+    completionHandler(UIBackgroundFetchResultNewData);
+  }
 }
 // [END receive_message]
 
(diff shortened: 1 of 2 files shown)


messaging/MessagingExampleSwift/AppDelegate.swift
```diff
@@ -91,6 +91,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     // Print full message.
     print(userInfo)
+
+    if completionHandler {
```
ianbarber: Does this work? Completion handler doesn't look nullable, and I've generally had to do != nil for these tests.

ianbarber:
LGTM bar swift q

messaging/MessagingExample/AppDelegate.m
```diff
@@ -125,6 +125,10 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
 
   // Print full message.
   NSLog(@"%@", userInfo);
+
+  if (completionHandler) {
```
revolter: Same here, as the documentation states: "the completion handler _must_ be called", so it wouldn't make sense to check it.

revolter:
Funny, I just wanted to create this PR too.

messaging/MessagingExample/AppDelegate.m
kroikie: Agreed, removed the check from both Obj C and Swift.

messaging/MessagingExampleSwift/AppDelegate.swift
kroikie: It does not, you are correct here Ian, however seems like an unnecessary check since this is called by the OS, the handler should always be available.

messaging/MessagingExample/AppDelegate.m
```diff
@@ -125,6 +125,8 @@ - (void)application:(UIApplication *)application didReceiveRemoteNotification:(N
 
   // Print full message.
   NSLog(@"%@", userInfo);
+
+  completionHandler(UIBackgroundFetchResultNewData);
```
revolter: @kroikie, Any specific reason for using this instead of `UIBackgroundFetchResultNoData`?

PR #195, opened 2016-12-15
Add missing completion handlers calls

revolter:
Related to #189


kroikie:
This looks good, thanks!

PR #197, opened 2016-12-15
update firebase ios quickstarts to modern objc syntax

iulukaya:
Change-Id: I188dd35bc99dc285e3c84327e2a21584609216ea


admob/AdMobExample/ViewController.m
```diff
@@ -79,7 +79,7 @@ - (void)interstitialDidDismissScreen:(GADInterstitial *)interstitial {
 }
 
 - (IBAction)didTapInterstitialButton:(id)sender {
-  if ([self.interstitial isReady]) {
+  if ((self.interstitial).isReady) {
```
revolter: The parentheses are superfluous. Xcode puts them because he doesn't know for sure if they're needed or not. It happened to me too in a real application and I went and removed them manually.

PR #204, opened 2016-12-24
Change UIAlertController style from action sheet to alert

ourguide:
Action sheet style did not work on ipad


firattemel:
In order for action sheet to work on iPad sourceView of the picker.popoverPresentationController should be set like this: 
`if let popoverController = picker.popoverPresentationController{

popoverController.sourceView = //Your sourceView to show the tooltip of the action sheet

}`

ourguide:
Thank you.
I know what you said. It is suggested that the application of the style is good.

PR #268, opened 2017-05-16
Readability improvement in config's README.md

sphippen:
The negative-heavy structure of the original paragraph made it difficult to comprehend, or at least more difficult than it needs to be.

For example https://twitter.com/Gemmbu/status/864422652198305796
(rough TL:"This English reads like the negation of a negation, hard to get")

The rewrite should be more straightforward.


morganchen12:
Thanks @sphippen!

PR #287, opened 2017-05-30
Call configure before registering w/ notifications

rsattar:
This helps us avoid a potential race condition where iOS could return an APNs token before Firebase (and FIRMessaging) has been configured, potentially not associating the APNs token with the FCM token.


rondinellimorais:
Where is the documentation for a new version?

**FirebaseApp** now is **FIRApp**

**Messaging.messaging().delegate** now is **FIRMessaging.messaging().remoteMessageDelegate**

**didRefreshRegistrationToken** not exists anymore

thanks

PR #353, opened 2017-10-16
Fix Firebase Snapshot parsing on swift 4 - xcode 9

tiagozoom:
Hi, I've found an issue with the Swift version of the DatabaseExample. The Snapshot parsing was not working properly because it couldn't transform the response into a [String:String] dictionary. The fix was rather simple, I just parsed the snap response into a [String:AnyObject] dictionary and included the starCount value in the response (it was not included so the number of stars of a given comment was always zero).

### DIFF
--- a/database/DatabaseExampleSwift/Post.swift
+++ b/database/DatabaseExampleSwift/Post.swift
@@ -34,17 +34,18 @@ class Post: NSObject {
   }
 
   init?(snapshot: DataSnapshot) {
-    guard let dict = snapshot.value as? [String: String] else { return nil }
-    guard let uid  = dict["uid"]  else { return nil }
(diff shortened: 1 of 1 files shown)


tiagozoom:
Ok, I've signed the CLA.

 Thanks!

tiagozoom:
Hi, I've found an issue with the Swift version of the DatabaseExample. The Snapshot parsing was not working properly because it couldn't transform the response into a [String:String] dictionary. The fix was rather simple, I just parsed the snap response into a [String:AnyObject] dictionary and included the starCount value in the response (it was not included so the number of stars of a given comment was always zero).

tiagozoom:
I signed it!

PR #362, opened 2017-10-23
match firebase namespace changes

brettkoonce:


morganchen12:
duplicates https://github.com/firebase/quickstart-ios/pull/352/files

Issue #363, opened 2017-10-23
MessagingExample compiling error


valeriyvan:
MessagingExample/AppDelegate.m:45:28: Property 'delegate' not found on object of type 'FIRMessaging *'

kroikie:
Hi @valeriyvan what version of the pod are you using? The current version is 2.0.5. Property `delegate` should be there, it is recommended to be used instead of `remoteMessageDelegate` which is [deprecated](https://github.com/firebase/firebase-ios-sdk/blob/master/Firebase/Messaging/Public/FIRMessaging.h#L308).

PR #364, opened 2017-10-23
Fixes MessagingExample compile error

valeriyvan:
Fixes MessagingExample compile error using remoteMessageDelegate property instead of delegate

Fixes issue #363


valeriyvan:
I have signed CLA.

morganchen12:
You're using an older version of Firebase. Run `pod update` and the errors will go away.

PR #371, opened 2017-10-27
Refactoring NSUserDefaults usage:

valeriyvan:
- get rid of not needed synchronize calls;
- using stringForKey instead of valueForKey.


morganchen12:
Thanks!

PR #377, opened 2017-11-01
Solves static analiser issue

valeriyvan:
Adds one more else case with exception to stop static analyses complaining about method possibly returning nil


morganchen12:
Thanks!

PR #387, opened 2017-11-14
Shortens CGRect creation

valeriyvan:


database/DatabaseExample/PostDataSource.m
```diff
@@ -42,7 +42,7 @@ - (NSInteger)tableView:(UITableView *)tableView
 }
 
 - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView {
-  UILabel *noDataLabel         = [[UILabel alloc] initWithFrame:CGRectMake(0, 0, tableView.bounds.size.width, tableView.bounds.size.height)];
+  UILabel *noDataLabel         = [[UILabel alloc] initWithFrame:(CGRect){.size = tableView.bounds.size}];
```
morganchen12: I think objc developers who don't have a strong C background are more used to reading the more verbose syntax using `CGRectMake`, so we should leave it that way.

morganchen12:
Thanks!

PR #389, opened 2017-11-15
update invites example

ulukaya:


invites/InvitesExample/AppDelegate.m
```diff
@@ -43,28 +43,25 @@ - (BOOL)application:(UIApplication *)application
...
-
-  return [[GIDSignIn sharedInstance] handleURL:url
-                             sourceApplication:sourceApplication
-                                    annotation:annotation];
+  // [START_EXCLUDE silent]
+  __weak AppDelegate *weakSelf = self;
+  // [END_EXCLUDE]
+  // Handle App Invite requests
+  return [FIRInvites handleUniversalLink:url
+                                      completion:^(FIRReceivedInvite * _Nullable receivedInvite,
```
morganchen12: fix indent

invites/InvitesExample/AppDelegate.m
```diff
@@ -43,28 +43,25 @@ - (BOOL)application:(UIApplication *)application
...
+                      sourceApplication:sourceApplication
+                                 annotation:annotation]) {
     return YES;
   }
-
-  return [[GIDSignIn sharedInstance] handleURL:url
-                             sourceApplication:sourceApplication
-                                    annotation:annotation];
+  // [START_EXCLUDE silent]
+  __weak AppDelegate *weakSelf = self;
```
morganchen12: This probably doesn't need to be a weak variable, but it's inconsequential to the snippets.

OleksiyA: Normally I would prefer to have weak var here. But taking into account that this is quickstart sample, so that developers can quickly start using the library, I agree with Morgan and prefer not to complicate this code without need.

PR #398, opened 2017-11-26
Refactors if statement into guard. Breaks retain circle.

valeriyvan:


invites/InvitesExampleSwift/SignInViewController.swift
```diff
@@ -37,15 +37,16 @@ class SignInViewController: UIViewController, GIDSignInDelegate, GIDSignInUIDele
...
-      // User Successfully signed in.
-      // TODO: Remove async after, when GIDSignIn is started getting called after dissmissVC
-      DispatchQueue.main.asyncAfter(deadline: .now() + 0.35) {
-        self.performSegue(withIdentifier: "SignedInScreen", sender: self)
-      }
+      return
+    }
+    // User Successfully signed in.
+    // TODO: Remove async after, when GIDSignIn is started getting called after dissmissVC
+    DispatchQueue.main.asyncAfter(deadline: .now() + 0.35) { [weak self] in
```
morganchen12: Weak self isn't required here, since the dispatch queue releases the block after executing it. Self doesn't own the dispatch queue since it's a global queue, so there is no retain cycle.

Please remove the weak self and following guard statement, otherwise the changes look good.

PR #400, opened 2017-11-27
Changes unusually looking `case _` to `default`.

valeriyvan:


morganchen12:
The pattern-matching functional folk see `case _:` very naturally as "match any pattern". On the other hand, the C-based language folk see `default` as more natural

Since this is a matter of style, I'm going to not merge this CL. People who adopt our sample code should change the style to suit their codebase, not ours.

PR #408, opened 2017-12-11
If FIRDynamicLink is not-null, Firebase Dynamics Links SDK did handled the link.

OleksiyA:


OleksiyA:
Follow up for https://github.com/firebase/quickstart-ios/pull/407

PR #442, opened 2018-03-21
Better solution for checking os version

nhnam:
The macro you defined won't work. Better check if that interface was defined for which os. Check that interface by Cmd+Click to see.

### DIFF
--- a/messaging/MessagingExample/AppDelegate.m
+++ b/messaging/MessagingExample/AppDelegate.m
@@ -19,11 +19,8 @@
 
 // Implement UNUserNotificationCenterDelegate to receive display notification via APNS for devices
 // running iOS 10 and above.
-#if __IOS_AVAILABLE
 @interface AppDelegate () <UNUserNotificationCenterDelegate>
 @end
-@end
-#endif
 
 // Copied from Apple's header in case it is missing in some cases (e.g. pre-Xcode 8 builds).
 #ifndef NSFoundationVersionNumber_iOS_9_x_Max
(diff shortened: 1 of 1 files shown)


morganchen12:
@nhnam the macros here check for specific availabilities of OS versions in order to be compatible with older versions of Xcode. Removing the version checks in the macros breaks that.

There is a point of improvement to be made in this code, though. Instead of checking OS versions, we should check directly for the availabilities of the APIs we're using; for example:

```swift
if ([UNUserNotificationCenter class] == nil) {
  // UNUserNotificationCenter isn't available, use iOS 9 notifications.
}
```
instead of
```swift
if (floor(NSFoundationVersionNumber) <= NSFoundationVersionNumber_iOS_9_x_Max) {
  // iOS 9 notifications
}
```

Can you update your PR to use these checks instead?

nhnam:
Hi @morganchen12 ,
Thanks for your feedback, I've update my PR, please review it then let me know your idea.
Thanks :)

messaging/MessagingExample/AppDelegate.m
```diff
@@ -15,15 +15,14 @@
...
 @import UserNotifications;
-#endif
 
 // Implement UNUserNotificationCenterDelegate to receive display notification via APNS for devices
 // running iOS 10 and above.
-#if defined(__IPHONE_10_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0
+#if __IOS_AVAILABLE
 @interface AppDelegate () <UNUserNotificationCenterDelegate>
 @end
+@end
```
morganchen12: Please remove the extra `@end` declaration.

The `__IOS_AVAILABLE` wrapper here is probably unnecessary, since the whole AppDelegate class depends on UIKit/iOS.

messaging/MessagingExample/AppDelegate.m
```diff
@@ -60,24 +59,22 @@ - (BOOL)application:(UIApplication *)application
...
-          }];
-      #endif
+        // For iOS 10 display notification (sent via APNS)
+        [UNUserNotificationCenter currentNotificationCenter].delegate = self;
+        UNAuthorizationOptions authOptions =
+        UNAuthorizationOptionAlert
+        | UNAuthorizationOptionSound
+        | UNAuthorizationOptionBadge;
+        [[UNUserNotificationCenter currentNotificationCenter] requestAuthorizationWithOptions:authOptions completionHandler:^(BOOL granted, NSError * _Nullable error) {
+        }];
```
morganchen12: nit: The project uses two-space indents

nhnam: I've just fixed it.
Thanks @morganchen12

PR #458, opened 2018-05-08
fix mlkit sample detect button doesn't work

d-date:
Connect IBAction to detect button to show detect menu.


morganchen12:
Thanks @d-date!

PR #504, opened 2018-08-01
Remove FirebaseNanoPB from linker flags.

ryanwilson:


ryanwilson:
Is there a reason we have to have these linker flags here in the Xcode project, or can we remove them? CocoaPods should take care of them.

ulukaya:
Yeah, they are not needed. I'll make a cleanup

PR #518, opened 2018-08-16
update a commented out code for being more correct

mylyuic:


morganchen12:
This needs to be changed here as well: 

https://github.com/firebase/quickstart-ios/blob/master/inappmessaging/InAppMessagingExample/AppDelegate.m#L30

mylyuic:
updated objective c code as well.

PR #536, opened 2018-09-11
Re-arranged navigation to fix one-way trip to PatternTabBarController

elilienstein:
When navigation issues don't get in the way, exploration is enhanced.


analytics/AnalyticsExampleSwift/PatternTabBarController.swift
```diff
@@ -51,9 +51,11 @@ class PatternTabBarController: UITabBarController {
   }
 
   @IBAction func unwindToHome(_ segue: UIStoryboardSegue?) {
-
-  }
-
+    if let nc = self.navigationController {
+    nc.dismiss(animated: true, completion: nil)
```
morganchen12: Please fix the indents here (two spaces each)

PR #544, opened 2018-09-15
Import missing 'FirebaseMessaging' packages

devnokiyo:
If you use the current sample code, a build error will occur.
ex)
```
Messaging.messaging().delegate = self
```
Use of unresolved identifier 'Messaging'

I think you need to import 'FirebaseMessaging' package.

### DIFF
--- a/messaging/MessagingExampleSwift/AppDelegate.swift
+++ b/messaging/MessagingExampleSwift/AppDelegate.swift
@@ -18,6 +18,7 @@ import UIKit
 import UserNotifications
 
 import Firebase
+import FirebaseMessaging
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {


morganchen12:
That's supposed to be included automatically in the `import Firebase` statement above it, but this is not the first time it's failed to do so. Does this build issue go away when you try to actually build the project, or does it cause the build to fail?

devnokiyo:
An error was displayed before building and it failed to build.
However, when I investigated it again, the problem was solved by executing the following command.
```
rm -rf ~/Library/Caches/CocoaPods/
rm -rf Pods
pod install
```
This makes import description unnecessary.
```
import FirebaseMessaging
```
Sorry for lack of confirmation. I'd like to "Close" if you can confirm the content.

morganchen12:
Yep, sounds good. If this becomes a significant usability issue, though, it may be worth adding this import anyway since it's harmless and can help avoid one more build issue.

Issue #568, opened 2018-10-24
Update to Swift 4.2


manuelsBTS:
Some features of this are deprecated.  Needs an update

morganchen12:
We're not updating temporarily since some of our internal testing infrastructure isn't compatible with the newest Xcode yet.

ulukaya:
https://github.com/firebase/quickstart-ios/commit/2c7f4497dc3a2539b06fc17ba168553e5494b934

PR #590, opened 2018-11-28
Update to swift 4.2

redcapua:
Update to swift 4.2 according to issue #568


morganchen12:
Thanks for this migration work @redcapua!

ulukaya:
This shouldn't be merged yet. Our test infra doesn't support Xcode 10 or Swift 4.2. There is an update I can push which checks for swift version and works for both 4.2 and 4. I've been holding off on it to move to 4.2 straight when our infra is updated. I should've commented earlier.

PR #592, opened 2018-11-29
Update ViewController.swift

egilmorez:
Fix typo "instange" > instance.

In other cases I've been able to directly edit a file without creating a full-fledged pull request . . . but that option doesn't seem to exist here.

Thanks for reviewing!


morganchen12:
Thanks gilmorez!

PR #609, opened 2018-12-19
Demonstration of WIP Firestore Codable

paulb777:
This is for demonstration only currently and not intended for merging.

This converts part of the Firestore quickstart to use the Codable implementation at https://github.com/firebase/firebase-ios-sdk/pull/2178.

Note that this is based on the original contribution from https://github.com/firebase/firebase-ios-sdk/pull/838 and the API requested by @wilhuff at https://github.com/firebase/firebase-ios-sdk/issues/627#issuecomment-366466081.

The next step is resolving this difference and converging.


paulb777:
Replaced by #787

PR #649, opened 2019-02-27
Enable Microsoft sign in.

Yue-Wang-Google:


authentication/README.md
```diff
@@ -21,6 +21,18 @@ Getting Started
 - Run the app on your device or simulator.
     - Select **Sign In** and select Google to begin.
 
+### Microsoft Sign In Setup
+- Go to the [Firebase Console](https://console.firebase.google.com) and navigate to your project:
+  - Select the **Auth** panel and then click the **Sign In Method** tab.
+  - Open panel on bottom right corner and click **Experiment Panel**.
+  - Enable **FirebaseAuthProvidersExperimentEnabled** experiment
```
samtstern: Are these going to be the public instructions?

authentication/AuthenticationExample/MainViewController.m
```diff
@@ -82,6 +83,7 @@
 
 @interface MainViewController ()
 @property(strong, nonatomic) FIRAuthStateDidChangeListenerHandle handle;
+@property(strong, nonatomic) FIROAuthProvider *microsoftProvider;
```
renkelvin: Seems no need to store it as a property.

authentication/README.md
Yue-Wang-Google: Just double checked with our TL and you are right, this is not needed when we launch it. I modified the instruction.

authentication/AuthenticationExample/MainViewController.m
Yue-Wang-Google: OAuthProvider is different than other providers. Others can sign in use class (static) method. OAuth sign in is an instance method. So an object needs to be generated and we need to store a pointer somewhere. Otherwise, when running the block call, it will find the weak self pointer is already deallocated.

renkelvin: SGTM.

samtstern:
Did this feature go out already on iOS and in the console?  I know it's not released on Android yet.

bojeil-google:
Please revert this. This is going out next thursday in the console/docs.

samtstern:
Reverted: https://github.com/firebase/quickstart-ios/pull/651

ulukaya:
Thanks for the catch Sam!

PR #655, opened 2019-03-06
Reenable "Enable Microsoft sign in."

Yue-Wang-Google:
Reverts firebase/quickstart-ios#651


Yue-Wang-Google:
The feature is launched for Firebase iOS SDK so we can roll the sample out.

PR #666, opened 2019-03-27
Remove Twitter sign in in Auth samples.

Yue-Wang-Google:
we can submit this first.


Yue-Wang-Google:
Hi, Ibrahim. we're going to deprecate TwitterAuth{Provider, Credential} in the next sdk release. Developers are encouraged to use OAuth instead. Do you think it's a good idea to remove Twitter sign in in the quickstart app?

ulukaya:
Hi, I think we should keep this till the next release as well. Removing it is breaking docs. I'll shortly rollback this change and roll-forward with next release.

PR #678, opened 2019-04-10
Change IPHONEOS_DEPLOYMENT_TARGET to 11.2

Chava:
Change IPHONEOS_DEPLOYMENT_TARGET to 11.2 (from 8.0, which caused issues in AdMob Swift tests)


morganchen12:
@googlebot ignoring CLA since Nikita works at google. LDAP nzadiriushko

PR #689, opened 2019-04-23
[5-2336000026166] RegEx condition not working for User property. Working example.

vegemanis:
**Setup:**
1. In the firebase dashboard, create a user property called AppVersion
2. In remote config, have the following parameter defined:
* Name: unsupported
  * Condition: User Property AppVersion contains regex ^2\.1\. -> true
  * default -> false
  
**Issue:** 
Setting the User Property to "2.10.0.0" in the application, returns true for unsupported on fetch even though the regex does not match. true should only be returned for "2.1.(anything)". The period is not being respected. The regex is valid RE2.

### DIFF
--- a/config/ConfigExample.xcodeproj/project.pbxproj
+++ b/config/ConfigExample.xcodeproj/project.pbxproj
@@ -3,10 +3,14 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 50;
+	objectVersion = 51;
 	objects = {
 
 /* Begin PBXBuildFile section */
+		00312B4FEE2D84BAFB635620 /* Pods_ConfigExample_ConfigExampleUITests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 49CEDA70E7583B8DAE18D797 /* Pods_ConfigExample_ConfigExampleUITests.framework */; };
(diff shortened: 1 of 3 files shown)


morganchen12:
Sorry for the slow response here, this PR slipped through my notification inbox. Is this still an issue on Firebase 6?

vegemanis:
Just tested with Firebase 6.2 and it's still an issue.

morganchen12:
Filed b/135531574 internally. I'll keep this PR updated.

vegemanis:
@morganchen12 can you let me know of the status of this internal bug?

morganchen12:
No update yet. I've pinged the bug for you, so hopefully it will get a response soon.

nhoctre:
config/ConfigExample.xcodeproj/project.pbxproj
config/ConfigExampleSwift/ViewController.swift

shehrozkhan88:
vegemanis:regexIssue #1089

vegemanis:
@shehrozkhan88 @morganchen12 Still seeing this issue with the latest remote config example and firebase 7.2. Code changes have been added in my branch.

shehrozkhan88 (approved):

morganchen12:
Hey, apologies for the slow response. I was only able to reproduce this with regex `^2.1.`, not `^2\.1\.`. Seems to be working as intended. Please re-open if that's not the case.

PR #716, opened 2019-05-30
 Replace TwitterKit with OAuth signin

Yue-Wang-Google:


authentication/AuthenticationExampleSwift/MainViewController.swift
```diff
@@ -382,6 +398,7 @@ class MainViewController: UITableViewController, GIDSignInUIDelegate {
     }
     // [END auth_listener]
     self.microsoftProvider = OAuthProvider.init(providerID:"microsoft.com");
+    self.twitterProvider = OAuthProvider.init(providerID:"twitter.com");
```
morganchen12: style nit: The `.init` before the parens is optional and should be omitted.

Yue-Wang-Google: Done.

PR #808, opened 2019-10-18
Add Sign In with Apple

morganchen12:
Ported to ObjC and pending merge once console and SDK support land.

Travis is currently failing on Podfile changes, which will be fixed once nonce support is released.

### DIFF
--- a/authentication/AuthenticationExampleSwift/MainViewController.swift
+++ b/authentication/AuthenticationExampleSwift/MainViewController.swift
@@ -840,7 +840,7 @@ class MainViewController: UITableViewController {
       }
 
       randoms.forEach { random in
-        if length == 0 {
+        if remainingLength == 0 {
           return
         }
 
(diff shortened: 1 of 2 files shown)


bojeil-google:
Looks good. Thanks @morganchen12!

authentication/AuthenticationExampleSwift/MainViewController.swift
```diff
@@ -786,5 +800,110 @@ class MainViewController: UITableViewController {
...
+    precondition(length > 0)
+    let charset: Array<Character> =
+        Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")
+    var result = ""
+    var remainingLength = length
+
+    while remainingLength > 0 {
+      let randoms: [UInt8] = (0 ..< 16).map { _ in
+        var random: UInt8 = 0
+        let errorCode = SecRandomCopyBytes(kSecRandomDefault, 1, &random)
```
kevinthecheung: Can we just generate L bytes and convert to a string? Does the character set matter if we're going to hash and output a hex string anyway?

morganchen12: afaik there's no string encoding where all possible values are valid characters. Using a character set is an easy way to avoid this.

PR #889, opened 2020-01-28
Functions - add URL scheme directly to Info.plist

maksymmalyhin:
A step to fix b/148415602


maksymmalyhin:
@paulb777 Originally I though the value is unconditionally set by the script which looked as an extra step. Now I found that the value may be overwritten in google3 in tests, so I'll probably close the PR after double-checking.

PR #898, opened 2020-02-04
Fix fork CI

paulb777:


paulb777:
Still works fine - since we don't try to run the tests - we only build.

PR #903, opened 2020-02-07
Update admob sample

morganchen12:
Do not merge yet!

### DIFF
--- a/admob/Podfile
+++ b/admob/Podfile
@@ -2,7 +2,7 @@ use_frameworks!
 platform :ios, '8.0'
 
 pod 'Firebase/Analytics'
-pod 'Firebase/AdMob'
+pod 'Google-Mobile-Ads-SDK'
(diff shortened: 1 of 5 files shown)


admob/Podfile
```diff
@@ -2,7 +2,7 @@ use_frameworks!
 platform :ios, '8.0'
 
 pod 'Firebase/Analytics'
-pod 'Firebase/AdMob'
+pod 'Google-Mobile-Ads-SDK'
```
paulb777: Since this is the Firebase quickstart, perhaps should still access the pod via the Firebase subspec?

morganchen12: This change was specifically requested in b/148497211, though I am lacking the context behind why.

paulb777: Context cleared up in b/148497211. Thanks.

PR #909, opened 2020-02-14
Remove explicit core from podfile

morganchen12:
See b/148639420 for rationale.

Strangely, the Installations sample doesn't have a Firebase.h in its Core headers after running `pod update`, so I was unable to remove the header import TODOs.

### DIFF
--- a/installations/InstallationsExample/AppDelegate.m
+++ b/installations/InstallationsExample/AppDelegate.m
@@ -18,7 +18,7 @@
 
 // TODO: Remove once FirebaseInstallations has released.
 #import <FirebaseInstallations/FIRInstallations.h>
-@import Firebase;
+@import FirebaseCore;
 
 @interface AppDelegate ()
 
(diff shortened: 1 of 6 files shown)


paulb777:
The Firebase.h omission might be intentional until FirebaseInstallations is ready for direct usage from non-Firebase pods. @maksymmalyhin ?

maksymmalyhin:
I thought we would add FirebaseInstallations to our umbrella Firebase pod when I was adding the TODO. Since we didn't do this eventually keepin a separate header should be ok.

Also looks like FirebaseInstallations is missing in `Firebase.h`, I'll send a PR to fix it.

As for the FirebaseInstallations as a separate product - I don't think there is something preventing developers from using it if they have a use case for it. I think, we can add Firebase/Installations subspec in one of the following releases so `pod 'Firebase/Installations'` will add `Firebase.h` to the project. WDYT?

morganchen12:
[Installations docs are public](https://firebase.google.com/docs/reference/swift/firebaseinstallations/api/reference/Classes), so it's available for developers to use directly, though no one in the past used InstanceID directly either unless they were trying to force an ID refresh.

PR #912, opened 2020-02-17
dict key must be type String when it is not defined

masatsch:
before: `userInfo[gcmMessageIDKey]`
aftrer: `userInfo["gcmMessageIDKey"]`

variable `gcmMessageIDKey` is not defined so it causes error.
I think you hope to set it as "gcmMessageIDKey".
You should also fix this website: https://firebase.google.com/docs/cloud-messaging/ios/receive


masatsch:
@googlebot I signed it!

masatsch:
sorry. I didn't find out you defined this variable.

PR #924, opened 2020-03-06
Add mfa support

renkelvin:


bojeil-google:
Hey @morganchen12, we will initially add documentation/snippets in [GCIP docs](https://cloud.google.com/identity-platform/). We will follow up and add to Firebase docs once this is available in the Firebase Console. One thing to really consider is to support this in FirebaseUI (Android, iOS and web). I think with sufficient demand, we can justify that.

morganchen12:
Makes sense, thanks for clarifying.

renkelvin:
@samtstern @morganchen12 Please take a look at the latest commit, which adds a new sign in entry "Email with MFA" and keeps the regular "Email" unchanged.

authentication/AuthenticationExampleSwift/MainViewController.swift
```diff
@@ -28,6 +28,8 @@ import GoogleSignIn
 import FBSDKCoreKit
 import FBSDKLoginKit
 
+var isMFAEnabled = false
```
morganchen12: nit: make this private

PR #940, opened 2020-03-23
Include GoogleService-Info.plist in Xcode projects

paulb777:
Simplify CI and make it clearer to users the need to add GoogleService-Info.plist to the project.


paulb777:
@morganchen12 I believe it should be covered by the instructions already in the README's:

[Add Firebase to your iOS Project](https://firebase.google.com/docs/ios/setup)

Note that there is nothing to replace. This PR does not add the file - only the entry in the project manifest which will cause the UI to be read until the setup instructions are followed.

morganchen12:
In that case we should add a "If necessary, delete the additional file reference to GoogleService-Info.plist" since dragging and dropping the file into Xcode will create a new reference instead of populating the existing (red) reference.

paulb777:
Oh, that's a good point. I shouldn't optimize the workflow for command line copies versus drag and drop. I'll look for another way to solve the CI issues.

PR #953, opened 2020-04-20
Replace objc code in Auth Swift QS with Swift

paulb777:
Replace AuthenticationExample/UIViewController+Alerts.[hm] with AuthenticationExampleSwift/UIViewController.swift  for the Swift Example. 

Inspired by https://stackoverflow.com/questions/61281909/type-firebaseauth-has-no-member-showtextinputprompt/61292023#61292023


authentication/AuthenticationExampleSwift/UIViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+                           completionBlock: @escaping ((Bool, String?) -> Void)) {
+    let prompt = UIAlertController(title: nil, message: message, preferredStyle: .alert)
+    let cancelAction = UIAlertAction(title: "Cancel", style: .cancel) { _ in
+      completionBlock(false, nil)
+    }
+    weak var weakPrompt = prompt
+    let okAction = UIAlertAction(title: "OK", style: .default) { _ in
+      if let strongPrompt = weakPrompt {
+        completionBlock(true, strongPrompt.textFields![0].text)
+      }
```
morganchen12, line 62: nit: this can be
```swift
if let text = weakPrompt?.textFields?.first?.text {
  completionBlock(true, text)
}
```

authentication/AuthenticationExampleSwift/UIViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+   @param completion Called after the spinner has been hidden.
+   */
+  func showSpinner(_ completion: (() -> Void)?) {
+    let alertController = UIAlertController(title: nil, message: "Please Wait...\n\n\n\n",
+                                            preferredStyle: .alert)
+    SaveAlertHandle.set(alertController)
+    let spinner = UIActivityIndicatorView(style: .whiteLarge)
+    spinner.color = UIColor(ciColor: .black)
+    spinner.center = CGPoint(x: alertController.view.bounds.size.width / 2,
+                             y: alertController.view.bounds.size.height / 2)
```
morganchen12, line 81: You can use `alertController.view.frame.midX` and `alertController.view.frame.midY` instead of dividing.

authentication/AuthenticationExampleSwift/UIViewController.swift
ryanwilson: So we don't want to have a completion handler if OK was pressed but there was no text?

paulb777: I think it's not possible for there to be nil text. Even if it's blank, it will still be the empty string.

authentication/AuthenticationExampleSwift/UIViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+   */
+  func showTextInputPrompt(withMessage message: String,
+                           completionBlock: @escaping ((Bool, String?) -> Void)) {
+    let prompt = UIAlertController(title: nil, message: message, preferredStyle: .alert)
+    let cancelAction = UIAlertAction(title: "Cancel", style: .cancel) { _ in
+      completionBlock(false, nil)
+    }
+    weak var weakPrompt = prompt
+    let okAction = UIAlertAction(title: "OK", style: .default) { _ in
+      if let strongPrompt = weakPrompt {
```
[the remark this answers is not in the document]
paulb777: Done.

paulb777:
Agreed about a bigger redesign needed. The UI of Signing in versus Create is very confusing about which of the two spots to add the id and password.

Will merge here as a pure Swift implementation of what used to be a mixed implementation.

PR #958, opened 2020-04-28
Add un-Combine sample code

morganchen12:
This code adds a screen that loads all the restaurants and then attempts to asynchronously fetch all of their reviews before combining it into one big table view.

DO_NOT_MERGE, this is just for testing, but review is welcome.

### DIFF
--- a/firestore/FirestoreExample/ShowAllViewController.swift
+++ b/firestore/FirestoreExample/ShowAllViewController.swift
@@ -1,5 +1,5 @@
 //
-//  Copyright (c) 2020 Google LLC.
+//  Copyright (c) 2020 Google LLC
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


firestore/FirestoreExample/ShowAllViewController.swift
```diff
@@ -0,0 +1,95 @@
+//
+//  Copyright (c) 2020 Google LLC.
```
paulb777, line 2: nit: no period see go/copyright

firestore/FirestoreExample/ShowAllViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+    return restaurants[section].name
+  }
+
+  override func numberOfSections(in tableView: UITableView) -> Int {
+    guard let restaurants = restaurants else { return 0 }
+    return restaurants.count
+  }
+
+  override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+    guard reviewCollections.count > section else { return 0 }
```
paulb777, line 91: >=

morganchen12: Count must be greater than section. For example if section is `0` and count is `0` the line below will crash with an out-of-bounds error.

PR #963, opened 2020-05-01
Fix ABT QS build failure with latest RC

paulb777:
Fix a new build failure with the latest RC - b/155290239

The better longer term fix would be to migrate the quickstart from InstanceID to Installations

Nightlies missed this failure because they automatically install all internal pods including Installations.


maksymmalyhin:
Thank you Paul!

PR #967, opened 2020-05-07
Remove deprecated mcs code

### LINKED ISSUE #932: MessagingRemoteMessage is deprecated.
https://github.com/firebase/quickstart-ios/blob/cef37defb946d04b9b7d0ec0414457b7200d95a6/messaging/MessagingExampleSwift/AppDelegate.swift#L173

`MessagingRemoteMessage` is deprecated: FCM direct channel is deprecated.
We should be using use APNs for downstream message handling.

chliangGoogle:
Fix #932 

Also return alert type in one of the display notification delegate to ensure notification can be displayed when app is in foreground. This might or might not be the reason of [#5519](https://github.com/firebase/firebase-ios-sdk/issues/5519) where users can't receive notification.

### DIFF
--- a/messaging/MessagingExampleSwift/AppDelegate.swift
+++ b/messaging/MessagingExampleSwift/AppDelegate.swift
@@ -134,7 +134,7 @@ extension AppDelegate : UNUserNotificationCenterDelegate {
     print(userInfo)
 
     // Change this to your preferred presentation option
-    completionHandler([])
+    completionHandler([[.alert, .sound]])
   }
 
   func userNotificationCenter(_ center: UNUserNotificationCenter,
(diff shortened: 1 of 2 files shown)


messaging/MessagingExampleSwift/AppDelegate.swift
```diff
@@ -134,7 +134,7 @@ extension AppDelegate : UNUserNotificationCenterDelegate {
     print(userInfo)
 
     // Change this to your preferred presentation option
-    completionHandler([])
+    completionHandler([[.alert, .sound]])
```
paulb777, line 137: Looks like the Objective-C sample has the same issue?

egilmorez:
Thanks for the heads-up Morgan!  It's true, I think we're not documenting "direct channel" message handling any longer.

messaging/MessagingExampleSwift/AppDelegate.swift
chliangGoogle: done.

PR #968, opened 2020-05-08
Remove expirationDuration

paulb777:
Proposing to remove expirationDuration from the Remote Config getting started flow to simplify usability:

- It prevents console changes from being fetched
- Even after changing the value, the previous run's fetch is still locked in
- I can't think of a use case where I want to call fetch but only fetch old values


paulb777:
cc: @karenyz

PR #971, opened 2020-05-21
Add framework tool

granluo:
This is to help xcproject add framework automatically.

### DIFF
--- a/scripts/add_framework_script.rb
+++ b/scripts/add_framework_script.rb
@@ -1,0 +1,41 @@
+#!/usr/bin/env ruby
+
+# Copyright 2020 Google LLC
+#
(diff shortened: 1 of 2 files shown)


scripts/add_framework_script.rb
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env ruby
+
+# Copyright 2020 Google
```
paulb777, line 3: Should be Google LLC for new files - see go/copyright

granluo:
Hmmm.. the travis tests seem to be failed, is that related to the change?

paulb777:
It looks like a network flake.  I restarted the travis job.

Either way, I can't see how this change would be related.

PR #974, opened 2020-05-26
Add framework testapp launcher

granluo:


granluo:
This is tested in the pr
https://github.com/firebase/firebase-ios-sdk/pull/5741/commits
The change is identical to the file with same name in the pr.
https://github.com/firebase/firebase-ios-sdk/pull/5741/files#diff-76fa1326038525c5b5e1fa7feb83e12a

Test is done in
https://github.com/firebase/firebase-ios-sdk/pull/5741/checks?check_run_id=751937451

config/ConfigExampleTests/AppTests.m
```diff
@@ -17,6 +17,7 @@
 #import <UIKit/UIKit.h>
 #import <XCTest/XCTest.h>
 
+
```
paulb777, line 20: Delete

PR #976, opened 2020-05-27
Update header file import

granluo:


config/ConfigExampleTests/AppTests.m
```diff
@@ -17,8 +17,8 @@
 #import <UIKit/UIKit.h>
 #import <XCTest/XCTest.h>
 
-#import "FIRApp.h"
-#import "FIRRemoteConfig.h"
+#import "FirebaseCore/FIRApp.h"
```
paulb777, line 20: Since these are public headers, use bracket syntax:

#import <FirebaseCore/FIRApp.h>

config/ConfigExampleTests/AppTests.m
```diff
@@ -27,7 +27,6 @@ @implementation AppTests
 
 
 - (void)testRemoteConfig {
-  [FIRApp configure];
```
paulb777, line 30: This may be causing the test failures

PR #978, opened 2020-05-28
Determine if a FIRApp exists before configuration.

granluo:


config/ConfigExampleTests/AppTests.m
```diff
@@ -27,7 +27,7 @@ @implementation AppTests
 
 
 - (void)testRemoteConfig {
-  [FIRApp configure];
+  if([FIRApp defaultApp] == nil){ [FIRApp configure]; };
```
paulb777, line 30: To make this scalable, move to the setup function.

Also, fix the style to:

if ([FIRApp defaultApp] == nil) { 
  [FIRApp configure]; 
}

paulb777: Here's a Swift example of what I mean https://github.com/firebase/firebase-ios-sdk/blob/master/FirebaseRemoteConfig/Tests/SwiftAPI/APITestBase.swift#L28

paulb777:
Please also move to the setUp function so it is shared for any additional tests that get added.

granluo:
Yep, FIRApp seems to be configured in appdelegate.
https://github.com/firebase/quickstart-ios/search?p=1&q=firapp+configure&unscoped_q=firapp+configure

will test the remoteconfig and populate to the rest quickstarts.

PR #994, opened 2020-06-24
Update Crashlytics run script to handle zip installs for swift quickstart.

granluo:
This is the same as [Update Crashlytics run script to handle zip installs](https://github.com/firebase/quickstart-ios/pull/902) but for Swift quickstart. Zip framework testing is putting zip in "${PROJECT_DIR}/run".


paulb777:
cc: @samedson

samedson:
Ah I must have missed Swift - thanks @granluo for the fix!

PR #1004, opened 2020-07-02
Update Add Framework script

granluo:
Auth testapp request libc++.dylib from /usr/lib and accelerate.framework from xcode frameworks dir. These two cannot be copied to another folder and imported to the xcode project. This update will enable this tool to import a single file.
The argument/options update will increase the readability.
The test is done for [remote config in firebase-ios-sdk framework testing workflow](https://github.com/firebase/firebase-ios-sdk/runs/828632641?check_suite_focus=true)

scripts/setup_quickstart_framework.sh from firebase-ios-sdk repo depends on this file and will be updated whlie this pr is submitted. The change for the tested pr is [here](https://github.com/firebase/firebase-ios-sdk/pull/5958/files).

### DIFF
--- a/scripts/add_framework_script.rb
+++ b/scripts/add_framework_script.rb
@@ -16,26 +16,55 @@
 
 require 'xcodeproj'
 require 'set'
-sdk = ARGV[0]
-target = ARGV[1]
-framework_dir = ARGV[2]
+require 'optparse'
+
+options = {}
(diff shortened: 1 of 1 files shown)


granluo:
This tool will be used [here](https://github.com/firebase/firebase-ios-sdk/pull/5962) in firebase-ios-sdk repo.

PR #1007, opened 2020-07-07
Update and add bundle.

granluo:


granluo:
This change is to add bundle to xcode project. It is tested in 
https://github.com/firebase/firebase-ios-sdk/pull/5982/files

PR #1015, opened 2020-07-17
Requesting review for SwiftUI Passwordless demo app

ncooke3:
Not sure where this project will live yet (will talk with dev rel) but for the sake of review, I've created a `swiftui` directory in this repo for this review (and to potentially house future `SwiftUI` projects?).

Please Review:
• `PasswordlessSwiftUIApp.swift`
• `ContentView.swift`

TODO:
- [x] README


swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,181 @@
...
+
+    }
+    .onOpenURL { (url) in
+      let link = url.absoluteString
+      if Auth.auth().isSignIn(withEmailLink: link) {
+        passwordlessSignIn(email: email, link: link) { (success) in
+          self.presentSheet = success
+        }
+      }
+    }
```
ncooke3, line 48: I did not call any Dynamic Links API's. Interestingly the above code works well regardless. If we use the dynamic link APIs to verify the incoming link is indeed a dynamic link, I think the approach below would work:

```swift
   // The above approach does not use the dynamic link APIs.
   .onOpenURL { (url) in
     DynamicLinks.dynamicLinks().handleUniversalLink(url) { (dynamicLink, error) in
       guard error == nil, let link = dynamicLink?.url?.absoluteString else {
         print(error!)
         return
       }

       if Auth.auth().isSignIn(withEmailLink: link) {
         passwordlessSignIn(email: email, link: link) { (success) in
           self.presentSheet = success
         }
       }
     }
   }
```

After chatting with DL team, it seems like calling `handleUniversalLink()` does handles reopening attribution and some analytics for the link. I'm unsure whether it is still necessary here though. Thoughts?

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,181 @@
...
+        if Auth.auth().isSignIn(withEmailLink: link) {
+          passwordlessSignIn(email: email, link: link) { (success) in
+            completion(success)
+          }
+        } else {
+          completion(false)
+        }
+      }
+    }
+}
```
ncooke3, line 181: Perhaps Firebase could start including view modifiers like this one, for example.
```swift
    /// It's also possible to refactor the code on line 41 by using a view modifier like:
    .handlePasswordlessLogin(forEmail: email) { (success) in
      self.presentSheet = success
    }
```

swiftui/PasswordlessSwiftUI/Podfile
```diff
@@ -0,0 +1,21 @@
+# Uncomment the next line to define a global platform for your project
+platform :ios, '14.0'
```
paulb777, line 2: Is 13.0 possible?

ryanwilson: 14 is required for the new `@main` and entire-SwiftUI  lifecycle (instead of AppDelegate)

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,181 @@
...
+            .frame(maxWidth: .infinity, alignment: .leading)
+        }
+        .padding(.leading)
+        
+        Image(systemName: "checkmark.circle")
+          .resizable()
+          .aspectRatio(contentMode: .fit)
+          .scaleEffect(0.5, anchor: /*@START_MENU_TOKEN@*/.center/*@END_MENU_TOKEN@*/)
+      }
+      .foregroundColor(.white)
```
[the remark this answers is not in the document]
ncooke3: I think white had looked the best on the solid orange background. Something about the black on the orange didn't look right. I'll add pictures when I commit so we can better compare!

morganchen12:
Directory structure LGTM. Once we have full coverage of SwiftUI samples we can move the SwiftUI samples to root and put the old samples in a subdirectory for better discoverability.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -12,72 +19,83 @@ import Firebase
...
+        passwordlessSignIn(email: email, link: link) { user, error in
+          isPresentingSheet = user?.isEmailVerified ?? false
         }
       }
     }
-    .sheet(isPresented: $presentSheet) {
+    .sheet(isPresented: $isPresentingSheet) {
       SuccessView(email: $email)
     }
+  }
```
ryanwilson: Just thought of a potential race condition here: if someone sends a sign in link, changes the email in the text field, then goes to their email and logs in, we'll show that they're logged in as the content of the text field, not the email address they actually used.

Can you think of some ways to get around this without changing the UI?

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -12,72 +19,83 @@ import Firebase
...
+    let actionCodeSettings = ActionCodeSettings()
+    actionCodeSettings.url = URL(string: "https://passwordlessswiftui.page.link/demo_login")
+    actionCodeSettings.handleCodeInApp = true
+    actionCodeSettings.setIOSBundleID(Bundle.main.bundleIdentifier!)
+
+    Auth.auth().sendSignInLink(toEmail: email, actionCodeSettings: actionCodeSettings)
   }
 
+  private func passwordlessSignIn(email: String, link: String,
+                                  completion: @escaping (_ user: User?, _ error: Error?) -> ()) {
```
ryanwilson, line 70: We should take advantage of the `Result` type here instead of providing two optionals in the callback.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -88,10 +106,13 @@ struct CustomStyledButton: View {
...
-    Button(title, action: action)
-    .padding()
-    .frame(maxWidth: .infinity)
-    .accentColor(.white)
+    Button(action: action) {
+      Spacer()
+      Text(title)
+        .padding()
+        .accentColor(.white)
+      Spacer()
```
ryanwilson, line 114: Does this work as is? It's unclear how these are assembled - is this an HStack, VStack, ZStack? I'm assuming it's just a `Group` but it's not clear.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
ncooke3: Update. When this code is triggered on a URL open, the `email` passed into `passwordlessSignIn` will be the current email typed in the textfield. `passwordlessSignIn` really just wraps Auth's "sign in with email and link method", which matches up the link and email and will return an error if the email and email the link was sent to don't match. 

So I dont think this race condition will happen because if a user enters email A and sends the link. Then changes the textfield to email B and then goes to open the link sent to email A. They will not be signed in when returning to the app because the link corresponding with email A no longer corresponds to the binded `email` property, which is now email B.

ncooke3: One could always keep the latest email the link was sent to when a user taps the "send sign in link" button. But then what if they send the link to 5 emails and only open the link on the 1,2,3, or 4th? So I think this might be ok to leave as it with the idea that whatever email is in the textfield should represent the email the users wants to login in with.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -67,11 +68,15 @@ struct ContentView: View {
...
     Auth.auth().signIn(withEmail: email, link: link) { result, error in
-      print(error ?? "Authentication was successful.")
-      let newUser = error == nil ? Auth.auth().currentUser : nil
-      completion(newUser, error)
+      if let error = error {
+        print("ⓧ Authentication error: \(error.localizedDescription).")
+        completion(.failure(error))
+      } else {
+        print("✔ Authentication was successful.")
+        completion(.success(Auth.auth().currentUser))
```
ryanwilson, line 78: This could be `result?.user` instead of getting the singleton again.

This doesn't need changing right now, just wanted to point out this is an example where our API could be improved - right now you're forced to return an optional User object (or you could force unwrap it) when there:should be only 2 use cases: successfully signed in, or an error.

If we wanted to handle this case fully, we could check if `result` is also `nil` and wrap our own custom error, returning it in the completion handler. Don't think it's fully necessary but thought it was a good learning opportunity.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -107,42 +112,43 @@ struct CustomStyledButton: View {
...
-        .padding()
-        .accentColor(.white)
-      Spacer()
+      HStack {
+        Spacer()
+        Text(title)
+          .padding()
+          .accentColor(.white)
+        Spacer()
+      }
```
ryanwilson, line 121: Nit: can you add a comment about why you're wrapping the text with two spacers in an H stack?

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -154,26 +160,29 @@ struct SuccessView: View {
...
-    }
+  static var previews: some View {
+    ContentView()
+  }
 }
 
 // MARK: Example Auth View Modifier
 
 extension View {
   func handlePasswordlessLogin(forEmail email: String,
```
ryanwilson, line 171: I would say if we're going to use this, let's keep it, otherwise let's delete it.

ryanwilson: I'm good if you want to use it - you may need to change the `String` argument to `Binding<String>` though (I think...)

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
ncooke3: yes- i understand!

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/PasswordlessSwiftUIApp.swift
```diff
@@ -0,0 +1,23 @@
...
+//  Copyright © 2020 Google LLC. All rights reserved.
+//
+
+import SwiftUI
+import Firebase
+
+@main
+struct PasswordlessSwiftUIApp: App {
+
+  init() {
```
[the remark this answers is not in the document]
peterfriese: For Firebase features that make use of AppDelegate swizzling, developers will probably need to use @UIApplicationDelegateAdaptor injection. Hence, I'd remove "recommended" from the comment.

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/PasswordlessSwiftUIApp.swift
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import SwiftUI
+import Firebase
+
+@main
+struct PasswordlessSwiftUIApp: App {
+  
+  /// With the removal of `AppDelegate`, it is recommended to configure Firbase here.
```
peterfriese, line 21: Typo: Firbase -> Firebase

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
```diff
@@ -0,0 +1,167 @@
...
+    .sheet(isPresented: $isPresentingSheet) {
+      SuccessView(email: email)
+    }
+  }
+
+  // MARK: - Firebase 🔥
+
+  private func sendSignInLink() {
+    let actionCodeSettings = ActionCodeSettings()
+    actionCodeSettings.url = URL(string: "https://passwordlessswiftui.page.link/demo_login")
```
peterfriese, line 63: I wonder if it might be a good idea to extract project-dependent constants and move them to a configuration struct that developers can then update with their project-specific values.

ncooke3: As far as project constants, I can think of the above URL and some of the button titles, textfield placeholders, etc. Are these items I should move to a configuration struct?

peterfriese: I was thinking mostly of the URL, which would also benefit from being decomposed into the prefix and short link. Button titles etc. should be dealt with proper i18n tools (which might be beyond the scope of this sample, but I'll leave it to @ryanwilson to comment on how stringent we are with i18n in our code samples).

ncooke3:
@ryanwilson Following up on our offline discussion yesterday about error handling... What are your thoughts on this latest commit? I experimented with a few things but thought this may be a nice way to handle/display errors?

swiftui/PasswordlessSwiftUI/PasswordlessSwiftUI/ContentView.swift
ryanwilson: Yep I think we can leave the i18n out for this - breaking the project constant out to a separate file SGTM, no other variables are needed there.

ryanwilson:
> @ryanwilson Following up on our offline discussion yesterday about error handling... What are your thoughts on this latest commit? I experimented with a few things but thought this may be a nice way to display errors using `Alert`?

I like it - thanks for doing so!

swiftui/PasswordlessSwiftUI/README.md
```diff
@@ -0,0 +1,42 @@
+# Using Firebase Auth's Email Link Login with SwiftUI
+This demo app demonstrates how to implement Firebase Auth's 
+[Email Link/Passwordless Authentication](https://firebase.google.com/docs/auth/ios/email-link-auth) 
+in an app created with SwiftUI. At WWDC 2020, Apple announced
+new lifeycle [APIs](https://developer.apple.com/documentation/swiftui/app) 
+that enabled developers to build apps entirely with 
+[SwiftUI](https://developer.apple.com/xcode/swiftui/). This project uses these 
+latest iOS14 APIs in conjunction with Firebase to provide a clear
```
ryanwilson, line 8: Nit: space between iOS and 14

swiftui/PasswordlessSwiftUI/README.md
```diff
@@ -0,0 +1,42 @@
+# Using Firebase Auth's Email Link Login with SwiftUI
```
ryanwilson, line 1: Nit: newline after the #

swiftui/PasswordlessSwiftUI/README.md
```diff
@@ -0,0 +1,42 @@
...
+This demo app demonstrates how to implement Firebase Auth's 
+[Email Link/Passwordless Authentication](https://firebase.google.com/docs/auth/ios/email-link-auth) 
+in an app created with SwiftUI. At WWDC 2020, Apple announced
+new lifeycle [APIs](https://developer.apple.com/documentation/swiftui/app) 
+that enabled developers to build apps entirely with 
+[SwiftUI](https://developer.apple.com/xcode/swiftui/). This project uses these 
+latest iOS14 APIs in conjunction with Firebase to provide a clear 
+example of how to build modern apps with Firebase and SwiftUI.
+
+# Getting Started
```
ryanwilson, line 11: Nit: newline after the #

swiftui/PasswordlessSwiftUI/README.md
```diff
@@ -0,0 +1,42 @@
...
+Getting this project up and running is similar to the other quickstarts 
+found in this repo. The following steps will clone the project, 
+install its dependencies, and open the `.xcworkspace` project in Xcode.
+```bash
```
ryanwilson, line 15: Nit: newline before this.

swiftui/PasswordlessSwiftUI/README.md
```diff
@@ -0,0 +1,42 @@
...
+$ git clone https://github.com/firebase/quickstart-ios.git
+$ cd swiftui/PasswordlessSwiftUI/
+$ pod install --repo-update
+$ open PasswordlessSwiftUI.xcworkspace
+```
+
+# Finished Demo
```
ryanwilson, line 22: Nit: newline after the #

PR #1020, opened 2020-07-28
Add account linking in auth qs and screenshots

ncooke3:
Adding support for account linking. Also added directory of qs screenshots. Made some slight refactoring/clarity improvements along the way.

TODO's:
- [x] Change up the `AuthProvider` enum


authentication/AuthenticationExample/Models/AuthProvider.swift
```diff
@@ -23,13 +23,28 @@ enum AuthProvider: String {
...
+    case .passwordless:
+      return "emailLink"
+    case .anonymous:
+      return "anonymous"
+    case .custom:
+      return "custom"
+    default:
+      return rawValue.lowercased().appending(".com")
+    }
+  }
```
morganchen12, line 47: It may make more sense to use the provider's id as the raw value of the enum and then have a method that transforms the enum into a display string, instead of having your display string back your model type and then transforming the display string into an ID.

ncooke3: This makes sense to me. I will adjust! 👍

PR #1021, opened 2020-07-28
Fix legacy analytics path link

ncooke3:


ncooke3:
@ulukaya  Since the legacy analytics qs is in a new directory, will we need to update this path link auth and config?

PR #1053, opened 2020-10-01
get token from messaging sdk

kroikie:
Demonstrate retrieving FCM registration token from the Messaging SDK.


morganchen12:
Looks like real build failures: https://travis-ci.org/github/firebase/quickstart-ios/jobs/731778931

The method is defined here: https://cs.opensource.google/firebase-sdk/firebase-ios-sdk/+/master:FirebaseMessaging/Sources/Public/FIRMessaging.h;l=392?q=FIRMessaging.h

Maybe the pods just need to be updated?

PR #1055, opened 2020-10-02
Fix notification name from Firebase 7 rename.

ryanwilson:


morganchen12 (approved):

paulb777:
Is there a way to fix to work for both Firebase 6 and 7?

ryanwilson:
> Is there a way to fix to work for both Firebase 6 and 7?

Not that I'm aware of - the type itself was renamed. I suppose we could use the raw string value but that won't help.

paulb777:
One hack would be to check for the existence of the FIRVersion class. Otherwise, we might need to hold off on this PR and disable the quickstart test until Firebase 7 releases.

paulb777:
Replaced by #1062 and #1063

PR #1056, opened 2020-10-02
update messaging API based on the Firebase 7

chliangGoogle:


morganchen12 (commented):
Changes LGTM. Paul brought up in a separate PR that there's an issue with merging these before Firebase 7 lands (doubly so in this case, since it updates a snippet on devsite). We need to resolve that issue before this PR can actually be merged.

chliangGoogle:
yeah, i'm holding on this to see how they resolve the other PR.

chliangGoogle:
@morganchen12 @paulb777  Somehow this passes the check. Do you think this is still an issue?

morganchen12:
It's likely because the delegate is ObjC under the hood and ObjC protocols don't enforce types for method parameters, only selector matching. Seems strange that Swift wouldn't raise at least a warning though--maybe because `String?` is a strict superset of `String`?

chliangGoogle:
Should this still be a blocker for submission? The other PR failed the CI, but this one not.

paulb777:
I'm fine to merge if CI stays green and there is an eventual plan to update the quickstart.

paulb777 (approved):

paulb777 (changes requested):
Oh wait, the CI wasn't triggered by the swift cchange

paulb777 (approved):
Right, it's only the firebase-ios-sdk change that is blocked. Sorry about my spinning.

PR #1075, opened 2020-10-28
Deduplicate classes in test targets

morganchen12:
I think this is causing the quickstart tests to fail without running any tests in https://github.com/firebase/firebase-ios-sdk/pull/5916.

Example: https://github.com/firebase/firebase-ios-sdk/pull/5916/checks?check_run_id=1323797486

### DIFF
--- a/database/DatabaseExample.xcodeproj/project.pbxproj
+++ b/database/DatabaseExample.xcodeproj/project.pbxproj
@@ -278,7 +278,6 @@
 				1073483520333B8C004A66D1 /* DatabaseExampleUITests */,
 				5F5A534D1ADE670C00F81DF0 /* Products */,
 				5F9961041AE0CF4F0034F503 /* Shared */,
-				74ABB871649589F6F33D45AF /* Pods */,
 			);
 			sourceTree = "<group>";
 			wrapsLines = 0;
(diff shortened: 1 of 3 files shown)


database/DatabaseExample.xcodeproj/project.pbxproj
```diff
@@ -278,6 +278,7 @@
 				1073483520333B8C004A66D1 /* DatabaseExampleUITests */,
 				5F5A534D1ADE670C00F81DF0 /* Products */,
 				5F9961041AE0CF4F0034F503 /* Shared */,
+				74ABB871649589F6F33D45AF /* Pods */,
```
paulb777, line 281: Delete

database/DatabaseExample.xcodeproj/project.pbxproj
```diff
@@ -355,6 +356,13 @@
 			path = DatabaseExampleTests;
 			sourceTree = "<group>";
 		};
+		74ABB871649589F6F33D45AF /* Pods */ = {
```
paulb777, line 359: Delete

paulb777 (commented):
You can also delete these by deleting the red items in the Xcode UI

database/DatabaseExample.xcodeproj/project.pbxproj
morganchen12: Done.

database/DatabaseExample.xcodeproj/project.pbxproj
morganchen12: Done.

paulb777:
I'm not seeing duplicated classes with Xcode 12 and shouldn't since Firebase is by default dynamic, so I suspect this change should be reverted since it [breaks the firebase-ios-sdk tests](https://github.com/firebase/firebase-ios-sdk/actions?query=workflow%3Adatabase).  The `testCreateNewValidPost` test seems to be flaky, but I'm not sure why.

morganchen12:
[This commit](28085ce6605095eb5418b596b81262fa35bad2f7) is required for the dynamic library targets. For some reason they're not copied into the test bundle, so the default loader path isn't picking them up.

Reverting this PR brings the dependency versions back down to 6.x.

PR #1081, opened 2020-11-12
Replace smart quotes in Podfile

paulb777:
Fix a `pod install` warning.


morganchen12 (approved):
Huh, weird that they were only closing quotes.

PR #1111, opened 2021-01-16
Exclude iOS 9 method from snippet

morganchen12:
I couldn't explicitly replace the method with the UNUserNotificationCenter method because it's already defined in the same file, but based on how our doc is structured it should be ok.


chliangGoogle (approved):
Sorry just saw this. Feel free to ping me next time if i'm not responding.

PR #1127, opened 2021-03-05
Creating Firestore FriendlyEats sample with SwiftUI

racheldoshcollins:
* Adds another target to the FirestoreExample project called FirestoreSwiftUIExample
* Writes data to Firestore, as the UIKit app did via the “Populate” button in the toolbar
* Reads that collection of restaurants from Firestore, displays in a list

Notes
* Setting up the app to run locally is the same process as the FirestoreExample
* TO DO: Auth, DetailView page (including adding a new rating), Filtering


firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
```diff
@@ -0,0 +1,21 @@
...
+//  FirestoreSwiftUIExample
+//
+
+import SwiftUI
+import Firebase
+
+@main
+struct FirestoreSwiftUIExampleApp: App {
+
+    init() {
```
peterfriese, line 12: Not sure about the QuickStarts, but Firebase iOS SDK has an official indent of 2 spaces. They also have a nice formatter script to save you the hassle of formatting the code yourself :-)

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+//  RestaurantListViewModel.swift
+//  FirestoreSwiftUIExample
+//
+
+import Combine
+import Firebase
+
+class RestaurantListViewModel: ObservableObject {
+
+    @Published var restaurants: [Restaurant] = []
```
peterfriese, line 11: Can be simplified to @Published var restaurants = [Restaurant]()

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+    private var db = Firestore.firestore()
+
+    func fetchData() {
+        db.collection("restaurants").addSnapshotListener { (querySnapshot, error) in
```
peterfriese, line 15: Don't forget to store the snapshot listener handle (and later release it).

peterfriese, line 15: Add `[weak self]` to the closure.

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+            if let error = error {
+                print("Error getting restaurants: \(error.localizedDescription)")
+                return
+            }
+
+            self.restaurants = querySnapshot?.documents.compactMap { document in
+                try? document.data(as: Restaurant.self)
+            } ?? []
```
peterfriese, line 23: Not sure the `?? []` is required. I use a similar pattern in some of my apps, and I've never had to do that.

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,121 @@
+//
+//  ContentView.swift
+//  FirestoreSwiftUIExample
+//
+
+import SwiftUI
+import Firebase
+
+struct ContentView: View {
+    let db = Firestore.firestore()
```
peterfriese, line 10: I'd generally refrain from directly accessing the backend from a view. Instead, let's try to educate our users to always use view models.

firestore/FirestoreSwiftUIExample/Views/RestaurantItemView.swift
```diff
@@ -0,0 +1,81 @@
...
+}
+
+struct RestaurantItemView_Previews: PreviewProvider {
+    static var previews: some View {
+        let data = Restaurant(id: .init(), name: "Pizza Place", category: "Pizza", city: "Austin", price: 2, ratingCount: 1, averageRating: 4, photo: Restaurant.imageURL(forName: "Place Place"))
+        RestaurantItemView(restaurant: data)
+    }
+}
+
+struct ImageThumbnail: View {
```
peterfriese, line 44: This might go into a separate file (esp. if it's reused elsewhere).

firestore/FirestoreSwiftUIExample/Views/RestaurantItemView.swift
```diff
@@ -0,0 +1,81 @@
...
+    var body: some View {
+        WebImage(url: imageURL)
+            .resizable()
+            .placeholder(Image(systemName: "photo"))
+            .aspectRatio(1, contentMode: .fill)
+            .frame(width: 100, height: 100, alignment: .leading)
+    }
+}
+
+struct Price: View {
```
peterfriese, line 56: This might go into a separate file (esp. if it's reused elsewhere).

Might also want to call it `PriceView`

peterfriese (changes requested):
Great to see a SwiftUI sample! I've left some comments.

firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
racheldoshcollins: Thanks! I was going to ask Morgan about this today :)

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
morganchen12: It's required in this case because `querySnapshot` is an optional.

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    func fetchData() {
+        db.collection("restaurants").addSnapshotListener { (querySnapshot, error) in
+            if let error = error {
+                print("Error getting restaurants: \(error.localizedDescription)")
+                return
+            }
+
+            self.restaurants = querySnapshot?.documents.compactMap { document in
+                try? document.data(as: Restaurant.self)
```
morganchen12, line 22: Consider logging the error instead of discarding it. Something like
```swift
do {
  return try document.data(as: Restaurant.self)
} catch let error {
  print(error)
  return nil
}
```

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,121 @@
...
+                ToolbarItem(placement: .navigationBarTrailing) {
+                    Button("Filter") {
+                        print(restaurantListViewModel.restaurants.count)
+                    }
+                }
+            }
+        }
+    }
+
+    func populate() {
```
morganchen12, line 40: Similar to Peter's comment above, this method should (if we're being clean with our architecture) live outside of the view layer. This wasn't the case when we wrote the original app, so you don't have to clean it up here, but a good place to put it would be on an extension of the Firestore class.

```swift
extension Firestore {

  func populate() {
    // ...
  }

}
```

firestore/FirestoreSwiftUIExample/Views/StarsView.swift
```diff
@@ -0,0 +1,34 @@
...
+
+    var body: some View {
+        HStack {
+            ForEach(1..<6) { i in
+                getStar(num: i, rating: rating)
+                    .frame(width: 12, height: 20)
+                    .foregroundColor(getStarColor(num: i, rating: rating))
+            }
+        }
+    }
```
morganchen12, line 19: Wow, this is way more concise than the old implementation :)

peterfriese: SwiftUI FTW :-)

firestore/FirestoreSwiftUIExample/Firestore+Extension.swift
```diff
@@ -0,0 +1,78 @@
+//
+//  Firestore+Extension.swift
+//  FirestoreSwiftUIExample
+//
+
+import Firebase
+
+extension Firestore {
```
peterfriese, line 8: Might be worth putting this into an "Extensions" folder.

firestore/FirestoreSwiftUIExample/Firestore+Extension.swift
```diff
@@ -0,0 +1,78 @@
...
+
+      let restaurantRef = collection.document()
+      do {
+        try restaurantRef.setData(from: restaurant)
+      } catch {
+        fatalError("Encoding Restaurant failed: \(error)")
+      }
+
+      let batch = self.batch()
+      //TODO: guard let user = Auth.auth().currentUser else { continue }
```
peterfriese, line 46: TODO or done?

firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  FirestoreSwiftUIExampleApp.swift
+//  FirestoreSwiftUIExample
+//
+
+import SwiftUI
+import Firebase
+
+@main
+struct FirestoreSwiftUIExampleApp: App {
```
peterfriese, line 10: I've been thinking about whether we should promote the more complete setup (including @ UIApplicationDelegateAdaptor).

@maksymmalyhin is working on a proposal to eliminate the need for method swizzling, and this would require developers to hook into the application delegate. See go/firebase-app-delegate-api-review for details.

So I think it might be useful to start socialising the idea of using @UIApplicationDelegateAdaptor.

What are people's thoughts - @morganchen12 @maksymmalyhin @racheldoshcollins ?

firestore/FirestoreSwiftUIExample/Model/Restaurant.swift
```diff
@@ -0,0 +1,115 @@
...
+    case city
+    case price
+    case ratingCount = "numRatings"
+    case averageRating = "avgRating"
+    case photo
+  }
+}
+
+extension Restaurant {
+
```
peterfriese, line 32: In some structs, you leave a blank line, it others you don't. Would recommend choosing one approach and then implement consistently.

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -0,0 +1,50 @@
+//
+//  RestaurantListViewModel.swift
```
peterfriese, line 2: Use copyright headers

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,53 @@
...
+  @ObservedObject var restaurantListViewModel = RestaurantListViewModel()
+
+  var body: some View {
+    NavigationView {
+      List(restaurantListViewModel.restaurants) { restaurant in
+        NavigationLink(destination: RestaurantDetailView()) {
+          RestaurantItemView(restaurant: restaurant)
+        }
+      }
+      .navigationBarTitle("Friendly Eats", displayMode: .inline)
```
peterfriese, line 19: Is there a particular reason you chose to use .inline?

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  ContentView.swift
+//  FirestoreSwiftUIExample
+//
+
+import SwiftUI
+import Firebase
+
+struct ContentView: View {
```
peterfriese, line 9: I think it might be nice to choose a more specific name, e.g. "RestaurantListView" or "RootView".

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+struct RestaurantDetailView: View {
+  var body: some View {
+    Text("Detail View Placeholder")
+  }
+}
+
+struct ContentView_Previews: PreviewProvider {
```
peterfriese, line 49: Update name of the preview provider according to the name you choose for the content view (see my comment above).

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
```diff
@@ -0,0 +1,53 @@
...
+          Button("Filter") {
+            print(restaurantListViewModel.restaurants.count)
+          }
+        }
+      }
+    }
+  }
+}
+
+struct RestaurantDetailView: View {
```
peterfriese, line 43: WIP?

firestore/FirestoreSwiftUIExample/Views/PriceView.swift
```diff
@@ -0,0 +1,33 @@
...
+
+  var body: some View {
+    Text(priceString(from: price))
+      .font(.footnote)
+      .foregroundColor(Color.gray)
+      .multilineTextAlignment(.trailing)
+  }
+
+  func priceString(from price: Int) -> String {
+    let priceText: String
```
peterfriese, line 19: In this particular case, you might be able to use `String(repeating: "$", count: price)`

https://developer.apple.com/documentation/swift/string/2427723-init

firestore/FirestoreSwiftUIExample/Views/StarsView.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import SwiftUI
+
+struct StarsView: View {
+  var rating: Int
+
+  var body: some View {
+    HStack {
+      ForEach(1..<6) { i in
```
peterfriese, line 13: Or `1...5` ?

peterfriese (commented):
Great work!

I spotted a couple of areas that looked like they're either still WIP or just need to be removed, but other than that this look really good.

firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
morganchen12: That's unnecessary for Firestore, which doesn't have any swizzling-dependent behavior. I'm ok with omitting that step in quickstarts that don't need it, but we will definitely want the complete setup as you're describing in Analytics or Messaging.

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
morganchen12: Yep, I told Rachel it'd be better to break up the quickstart into multiple PRs, so it's ok to leave this as a stub.

morganchen12 (approved):
LGTM, but please address Peter's comments before merging. Thanks!

firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
racheldoshcollins: The existing UIKit version of this app uses Auth, however; I just haven't gotten around to including Auth in the SwiftUI version yet. Does Auth have swizzling-dependent behavior? In that case, the more complete setup may be the way to go, but we can delay updating it until I add Auth to this quickstart (which isn't a high priority).

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
racheldoshcollins: Ah, thanks, this slipped my mind. I'll copy over from the UIKit version.

firestore/FirestoreSwiftUIExample/Views/PriceView.swift
racheldoshcollins: Cool! Will do

firestore/FirestoreSwiftUIExample/Views/StarsView.swift
racheldoshcollins: Ha yep

firestore/FirestoreSwiftUIExample/Firestore+Extension.swift
racheldoshcollins: Ah, I marked these at TODO because I haven't set up Auth. I'll remove for this PR.

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
racheldoshcollins: Just to match the UIKit app, but looking at it again, I'll remove so that the title has a more modern look

firestore/FirestoreSwiftUIExample/Views/ContentView.swift
racheldoshcollins: I added a "WIP" comment

firestore/FirestoreSwiftUIExample/FirestoreSwiftUIExampleApp.swift
peterfriese: As far as I can tell, Auth doesn't use swizzling. At least I haven't run into any issues so far when using the lighter weight SwiftUI + Firebase setup.

Agree with Morgan that it's fine to go with the light-weight (i.e. init()-only) approach if we don't need a full AppDelegate.

peterfriese:
Yay!

PR #1133, opened 2021-03-25
Adding DetailView SwiftUI to firestore sample

racheldoshcollins:
Adds RestaurantDetailView to displays reviews for an individual restaurant.

I successfully deintegrated this time :)

### DIFF
--- a/firestore/FirestoreExample.xcodeproj/project.pbxproj
+++ b/firestore/FirestoreExample.xcodeproj/project.pbxproj
@@ -31,8 +31,12 @@
 		8E4C637825EEF4CA001678A1 /* RestaurantListViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E4C637725EEF4CA001678A1 /* RestaurantListViewModel.swift */; };
 		8E4C63B225F05E76001678A1 /* StarsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E4C63B125F05E76001678A1 /* StarsView.swift */; };
 		8E4C63D825F6D641001678A1 /* Firestore+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E4C63D725F6D641001678A1 /* Firestore+Extension.swift */; };
-		8E4C63DE25F6D879001678A1 /* ImageThumbnailView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8E4C63DD25F6D879001678A1 /* ImageThumbnailView.swift */; };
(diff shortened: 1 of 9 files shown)


firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
```diff
@@ -47,12 +47,15 @@ class RestaurantListViewModel: ObservableObject {
         guard let self = self else { return }
         self.restaurants = documents.compactMap { document in
           do {
-            return try document.data(as: Restaurant.self)
+            var restaurant = try document.data(as: Restaurant.self)
+            restaurant?.reference = document.reference
```
morganchen12, line 51: There's a bit of trickiness going on here where the document reference is supported by the decoder, but it won't decode the document reference into a parameter of type `DocumentReference?` due to the optional. But we can't make the document reference nonnull on `Restaurant`, because the restaurant values we're working with don't necessarily exist on the server. Adding the optional document reference parameter directly to the `Restaurant` type increases its flexibility, but at the cost of increased complexity--when we do need to use the `reference` parameter, we can't be sure of its existence without examining the value directly, and when we don't, it requires extra cruft in initialization (such as in this example).

Fortunately, with granular types and generics, we can use a separate type to represent a context that we can drop the restaurant into. In this case, the context is 'a value that has a corresponding Firestore document', which we can represent like so:

```swift
struct FirestoreValue<T: Codable, Identifiable>: Identifiable {
  let value: T
  let snapshot: QueryDocumentSnapshot

  init(snapshot: QueryDocumentSnapshot) throws {
    self.snapshot = snapshot
    self.value = try snapshot.data(as: T.self)
  }
}
```

This way you can wrap a `Restaurant` in a `FirestoreValue` whenever it's instantiated as a result of a Firestore query, which lets you separate the `Restaurant`s that were created on the client from the ones that came from the server and allows you to access the Firestore bits directly when you need them. Since Firestore-backed `Restaurant`s and non-Firestore-backed `Restaurant`s are now representable as distinct types (`FirestoreValue<Restaurant>` vs `Restaurant`), we can use them in method parameters to explicitly specify whether or not we need to access the document reference, and the compiler will throw type errors if we're unable to.

Making this change is optional, since it'll require going back and refactoring some infrastructural code, so don't feel like you have to refactor this unless you're in the mood to wrangle the type system.

morganchen12 (approved):
LGTM, some nits and some food for thought.

firestore/FirestoreSwiftUIExample/ViewModels/RestaurantListViewModel.swift
racheldoshcollins: Great info. Thanks for the explanation. I'll give it a shot!

PR #1136, opened 2021-04-02
docs(functions): fix README's StackOverflow link

rosariopfernandes:
The StackOverlow link in the functions quickstart's README file was mistakenly pointing to Firebase Crash Reporting. This PR should point it to the correct url.


ncooke3 (approved):
Thanks Rosário!

PR #1143, opened 2021-04-20
Remove instanceID from abtesting quickstart app

chliangGoogle:
A/B Test no longer depends on InstanceID, remove IID check in quickstart app. Replace with FIS ID check.


morganchen12 (approved):
Thanks!

Issue #1099, opened 2020-12-11
Firestore Auto-update not happening


paulb777:
How is the auto-update configured?  It looks like it's not working for Firestore:

<img width="1016" alt="Screen Shot 2020-12-11 at 9 24 56 AM" src="https://user-images.githubusercontent.com/73870/101934757-e0571700-3b92-11eb-9485-0309d1af3d78.png">

morganchen12:
@ulukaya can you take a look at this?

ulukaya:
My script just finds "Podfile"s and updates them. https://github.com/GoogleCloudPlatform/repository-gardener/blob/master/use-latest-deps-ios.sh#L67
Somehow the latest Podfile update stopped it. I'll check what's going on.

ulukaya:
I don't see any errors but looks like it got timed out. I updated manually for now, that should make the next iteration work. I'll investigate further if it gets stuck again.

ulukaya:
https://github.com/firebase/quickstart-ios/commit/92f82d085c829271c8b2a812cd742e88c160bdae

paulb777:
Reopening since this is still a problem as can be seen from repo's main page's directories' last update messages.

morganchen12:
The culprit seems to be gRPC-C++:

```
Installing gRPC-C++ (1.28.2)
[!] /bin/bash -c 
set -e
sed -E -i '' 's;#include <openssl/(.*)>;#if COCOAPODS==1\
  #include <openssl_grpc/\1>\
#else\
  #include <openssl/\1>\
#endif;g' $(find src/core -type f \( -path '*.h' -or -path '*.cc' \) -print | xargs grep -H -c '#include <openssl_grpc/' | grep 0$ | cut -d':' -f1)
    find src/core/ third_party/upb/ -type f \( -name '*.h' -or -name '*.c' -or -name '*.cc' \) -print0 | xargs -0 -L1 sed -E -i'.grpc_back' 's;#include "upb/(.*)";#if COCOAPODS==1\
  #include  "third_party/upb/upb/\1"\
#else\
  #include  "upb/\1"\
#endif;g'
    find src/core/ third_party/upb/ -type f -name '*.grpc_back' -print0 | xargs -0 rm
[...]

PR #1146, opened 2021-04-28
Update all the Podfile.lock's

paulb777:
For Firebase 7.11.0 and GoogleUtilities 7.4.0


maksymmalyhin (approved):
LGTM

maksymmalyhin:
> Isn't there something like dependency bot that should be able to update these for us?

Looks like it hasn't been run [since January](https://github.com/firebase/quickstart-ios/pull/1116)...

paulb777:
Yep. It was initially partially broken #1099 and now seems completely broken.

PR #1158, opened 2021-05-04
Update quickstart to align with quickstart-js

morganchen12:
See https://github.com/firebase/quickstart-js/blob/master/functions/functions/index.js#L95


paulb777 (approved):

paulb777:
This doesn't build. See https://github.com/firebase/firebase-ios-sdk/issues/8022

PR #1160, opened 2021-05-05
Fix build breakage

morganchen12:
Breakage introduced in #1159.


paulb777:
Swift test needs to be added to .travis.yml

paulb777 (approved):
Thanks!

PR #1161, opened 2021-05-07
Move to GHA

morganchen12:


.github/workflows/admob.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+        matrix:
+          swift-suffix: ['', 'Swift']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@master
+      - name: Build
+        run: |
+          cd admob
```
paulb777, line 29: Issues are generally easier to diagnose for with separate named steps for the commands.  Especially to distinguish between setup and running.

.github/workflows/admob.yml
```diff
@@ -0,0 +1,39 @@
...
+
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  # This workflow contains a single job called "build"
+  build:
+    name: build
+    runs-on: macOS-latest
+    strategy:
+        matrix:
+          swift-suffix: ['', 'Swift']
```
paulb777, line 22: I would prefer to build Swift as an additional the run step. Both matrix elements have about 10 minutes of exact duplicate steps and a few seconds different.

.github/workflows/admob.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    steps:
```
paulb777, line 24: Also consider building and testing the test targets.

paulb777 (approved):
All of my comments can be considered optional or future items to address.

Thanks for doing this!

.github/workflows/admob.yml
morganchen12: This should be handled automatically by `scripts/test.sh`, but it looks like secrets always fail to install. Is this script intended to be used from firebase-ios-sdk's checks?

.github/workflows/admob.yml
morganchen12: Done.

.github/workflows/admob.yml
morganchen12: The tests fail without credentials present due to some header stuff that's used by firebase-ios-sdk. I'll merge the PR as-is and we can pursue adding the test targets separately.

paulb777: SGTM

PR #1168, opened 2021-05-12
Move secrets to GHA

morganchen12:


paulb777 (approved):
Does testing need to be turned on for PRs to catch in pre-submit instead of post-submit?

PR #1172, opened 2021-05-13
Delete Quickstarts for products no longer part of Firebase

paulb777:


morganchen12 (approved):
The ML quickstarts have scripts that could be removed from the [scripts directory](https://github.com/firebase/quickstart-ios/tree/master/scripts/install_prereqs) as well.

paulb777:
Looks like they're already gone

```
~/qs/210428/quickstart-ios (pb-delete) $ git grep mlvision
~/qs/210428/quickstart-ios (pb-delete) $ git grep smartreply
~/qs/210428/quickstart-ios (pb-delete) $ git grep -i smartreply
```

PR #1190, opened 2021-06-22
Pin FirebaseUI version and fix breakage

morganchen12:
Fixes b/191721090.


paulb777 (approved):
LGTM on green. 

Would you push an 11.1.0 with https://github.com/firebase/FirebaseUI-iOS/pull/968, update the Podfile.locks and reconfirm

PR #1191, opened 2021-06-23
Fix problem of non-asking users permission

jeyremy:
This fixes an issue were iOS don't ask users permissions and by consequence not generating the token for apns / FCM...


jeyremy:
@googlebot I signed it!

jeyremy:
I did not write details but here it is : 
IN SWIFT : 
if the line application.registerForRemoteNotifications() is called at the end of method : 
application(_ application: UIApplication,  didFinishLaunchingWithOptions ) 
the pop up "do you want notifications" is NEVER showed to the user...

chliangGoogle (changes requested):
I'm not sure if the order matters. If it does, it means there's a bug in apple's API.

morganchen12:
Closing this since it's staled, @jeyremy let me know if you'd like to re-open it and continue the discussion.

PR #1198, opened 2021-06-28
Separate SPM and CocoaPods samples

morganchen12:


scripts/test_legacy.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
paulb777, line 1: Would be better to add a flag to set a variable to the existing script for LEGACY.

paulb777 (approved):

scripts/test_legacy.sh
morganchen12: Done.

PR #1201, opened 2021-07-01
Add abtesting installation auth token usage

jrcrespoh:
Add installation auth token logging to abtesting quickstart sample and accompanying instructions.

### DIFF
--- a/abtesting/ABTestingExample/ViewController.swift
+++ b/abtesting/ABTestingExample/ViewController.swift
@@ -62,6 +62,14 @@ class ViewController: UIViewController, UITableViewDataSource {
 
   override func viewWillAppear(_ animated: Bool) {
     super.viewWillAppear(animated)
+    Installations.installations().authTokenForcingRefresh(true) { (token, error) in
+      if let error = error {
+        print("Error fetching token: \(error)")
+        return
+      }
+      guard let token = token else { return }
+      print("Installation auth token: \(token.authToken)")
+    }
     RemoteConfig.remoteConfig().fetch(withExpirationDuration: 0) { (status, error) in
       if let error = error {
         print("Error fetching config: \(error)")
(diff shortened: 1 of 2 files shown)


paulb777 (approved):
Thanks! LGTM - good to merge with one other approval.

abtesting/ABTestingExample/ViewController.swift
```diff
@@ -62,6 +62,14 @@ class ViewController: UIViewController, UITableViewDataSource {
 
   override func viewWillAppear(_ animated: Bool) {
     super.viewWillAppear(animated)
+    Installations.installations().authTokenForcingRefresh(true) { (token, error) in
+      if let error = error {
+        print("Error fetching token: \(error)")
+        return
+      }
+      guard let token = token else { return }
+      print("Installation auth token: \(token.authToken)")
```
ncooke3, line 71: Implementing handlers with two nullable values (token & error) can be awkward in Swift because you have to unwrap both values. If you're interested, take a look at [result types](https://www.swiftbysundell.com/basics/result/), which were created to make this easier. While this API doesn't support result type (though maybe it should 🤔 ), here's another way of doing this to consider (sharing in case you happen to prefer it more):
```swift
if let error = error {
 // ...
} else if let token = token {
 // ...
}
```

ncooke3 (approved):
LGTM– nice work! I added an optional suggestion.

abtesting/ABTestingExample/ViewController.swift
jrcrespoh: Thanks for the pointer! I'm thinking of keeping it the same just to stay relatively consistent with https://firebase.google.com/docs/ab-testing/abtest-config#validate_your_experiment_on_a_test_device and https://firebase.google.com/docs/projects/manage-installations#retrieve-fis-token.

PR #1207, opened 2021-07-02
Fix race in Firestore UI Tests

paulb777:
Don't try to sign in simultaneously from multiple tests.

Fix https://github.com/firebase/firebase-ios-sdk/issues/8159

### DIFF
--- a/firestore/FirestoreExampleUITests/FirestoreUITest.m
+++ b/firestore/FirestoreExampleUITests/FirestoreUITest.m
@@ -28,24 +28,28 @@ static NSString *const testPassword = @"Test123";
 @interface FirestoreUITest : XCTestCase
 @end
 
-@implementation FirestoreUITest {
-  XCUIApplication *_app;
+static XCUIApplication *sApp;
+
+@implementation FirestoreUITest
+
++ (void)setUp {
+  sApp = [[XCUIApplication alloc] init];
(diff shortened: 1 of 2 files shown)


firestore/FirestoreExampleUITests/FirestoreUITest.m
```diff
@@ -28,24 +28,28 @@
...
+static XCUIApplication *sApp;
 
-- (void)setUp {
-  [super setUp];
-  _app = [[XCUIApplication alloc] init];
-  [_app launch];
+@implementation FirestoreUITest
+
++ (void)setUp {
+  sApp = [[XCUIApplication alloc] init];
```
ncooke3, line 36: Im curious, what is the difference here and why does that change the behavior? Before it was an ivar (correct?) and now it's a `static` what? (it's not a property right?)

paulb777: The key change is moving the signin flow from the setup instance method that runs for every test to the setup class method that runs once for all the tests.

An alternative fix would have been to add an `@synchronized(self)` around the signin check and call.

ncooke3: Ahh, makes sense now, thanks. I did not notice the transition from instance to class method 😅

PR #1213, opened 2021-07-09
Add pull-down refreshing functionality on iOS 15 to abtesting [SwiftUI]

jrcrespoh:
Update app config from RemoteConfig using refreshable View modifier for iOS 15. Modularize ContentView with use of new struct BasicList. Add dynamic UI tests for refreshing functionality.

Any suggestions on how/whether to implement async/await functionality for this Quickstart?

### DIFF
--- a/abtesting/Shared/ContentView.swift
+++ b/abtesting/Shared/ContentView.swift
@@ -29,24 +29,30 @@ struct ContentView: View {
   var body: some View {
     NavigationView {
       VStack {
-        List(data, id: \.title) { item in
-          VStack(alignment: .leading) {
-            Text(item.title)
-            Text(item.subtitle)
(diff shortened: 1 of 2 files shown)


paulb777 (approved):
For async/await, it might be possible to update the `updateFromRemoteConfig` function to use the async versions of `fetch` and `activate` or even `fetchAndActivate` instead.

morganchen12 (approved):

jrcrespoh:
> For async/await, it might be possible to update the `updateFromRemoteConfig` function to use the async versions of `fetch` and `activate` or even `fetchAndActivate` instead.

Do you think it would be best to just include such changes in this PR or make a separate one?

paulb777:
Separate. This one is good to go as is.

PR #1214, opened 2021-07-09
Add async/await functionality to abtesting [SwiftUI]

jrcrespoh:
Use updateFromRemoteConfigAsync when using iOS 15 pull-down-to-refresh functionality.

Where else should async/await functionality be used? Should the app log something special to showcase async/await?

### DIFF
--- a/abtesting/Shared/AppConfig.swift
+++ b/abtesting/Shared/AppConfig.swift
@@ -57,6 +57,31 @@ class AppConfig: ObservableObject {
     }
   }
 
+  @available(iOS 15, *)
+  func updateFromRemoteConfigAsync() async {
+    let remoteConfig = RemoteConfig.remoteConfig()
+    do {
+      let status = try await remoteConfig.fetch(withExpirationDuration: 0)
+      print("Config fetch completed with status: \(status.debugDescription)")
+      do {
+        let changed = try await remoteConfig.activate()
+        let value = remoteConfig["color_scheme"].stringValue ?? "nil"
+        if changed {
+          print("Remote Config changed to: \(value)")
+          DispatchQueue.main.async {
+            self.colorScheme = ColorScheme(value)
+          }
+        } else {
+          print("Remote Config did not change from: \(value)")
+        }
+      } catch {
+        print("Error activating config: \(error)")
+      }
+    } catch {
+      print("Error fetching config: \(error)")
+    }
+  }
+
   @objc func printInstallationAuthToken() {
     Installations.installations().authTokenForcingRefresh(true) { token, error in
       if let error = error {
--- a/abtesting/Shared/AppConfig.swift
+++ b/abtesting/Shared/AppConfig.swift
@@ -57,30 +57,32 @@ class AppConfig: ObservableObject {
     }
   }
 
-  @available(iOS 15, *)
-  func updateFromRemoteConfigAsync() async {
-    let remoteConfig = RemoteConfig.remoteConfig()
-    do {
-      let status = try await remoteConfig.fetch(withExpirationDuration: 0)
-      print("Config fetch completed with status: \(status.debugDescription)")
+  #if swift(>=5.5)
+    @available(iOS 15, *)
+    func updateFromRemoteConfigAsync() async {
+      let remoteConfig = RemoteConfig.remoteConfig()
       do {
-        let changed = try await remoteConfig.activate()
-        let value = remoteConfig["color_scheme"].stringValue ?? "nil"
-        if changed {
-          print("Remote Config changed to: \(value)")
-          DispatchQueue.main.async {
-            self.colorScheme = ColorScheme(value)
+        let status = try await remoteConfig.fetch(withExpirationDuration: 0)
+        print("Config fetch completed with status: \(status.debugDescription)")
+        do {
+          let changed = try await remoteConfig.activate()
+          let value = remoteConfig["color_scheme"].stringValue ?? "nil"
+          if changed {
+            print("Remote Config changed to: \(value)")
+            DispatchQueue.main.async {
+              self.colorScheme = ColorScheme(value)
+            }
+          } else {
+            print("Remote Config did not change from: \(value)")
           }
-        } else {
-          print("Remote Config did not change from: \(value)")
+        } catch {
+          print("Error activating config: \(error)")
         }
       } catch {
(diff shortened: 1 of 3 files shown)


abtesting/Shared/AppConfig.swift
```diff
@@ -57,6 +57,31 @@ class AppConfig: ObservableObject {
     }
   }
 
+  @available(iOS 15, *)
+  func updateFromRemoteConfigAsync() async {
+    let remoteConfig = RemoteConfig.remoteConfig()
+    do {
```
jrcrespoh, line 63: Is there a better way of dealing with the nested do-catch statements?

morganchen12: No, this is just how error handling in Swift is.

abtesting/Shared/ContentView.swift
```diff
@@ -31,7 +31,7 @@ struct ContentView: View {
       VStack {
         if #available(iOS 15, *) {
           BasicList(data: data).refreshable {
-            appConfig.updateFromRemoteConfig()
+            await appConfig.updateFromRemoteConfigAsync()
```
morganchen12, line 34: Nice and succinct!

morganchen12 (approved):

abtesting/Shared/AppConfig.swift
paulb777: You could make the function throws, eliminate all the do, catches, and then do the do catch in the caller, but I think this is fine for this case other than my other comment.

abtesting/Shared/AppConfig.swift
```diff
@@ -57,6 +57,31 @@ class AppConfig: ObservableObject {
...
+      let status = try await remoteConfig.fetch(withExpirationDuration: 0)
```
paulb777, line 64: Consider using the fetchAndActivate API instead of the separate fetch and activate.

paulb777:
The test is failing because of missing #if swift 5.5 checks. See https://github.com/firebase/firebase-ios-sdk/pull/8289 for examples.

Before fixing that, it would be good to investigate why the CI is succeeding despite the error.

paulb777:
We might want to make a subset of https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh#L111 with the `RunXcodebuild()` function in the scripts directory of this repo as well.

paulb777:
You should be able to repro the CI build failures with Xcode 12.

jrcrespoh:
> Before fixing that, it would be good to investigate why the CI is succeeding despite the error.

@morganchen12 @paulb777 I am not very sure, but my current understanding is that while xcodebuild is failing, xcpretty is exiting successfully, but that seems to contradict other scripts which do something like `(xcodebuild "$@" | xcpretty) || EXIT_STATUS="$?"`...

Also, is there a better way of debugging CI issues instead of committing changes and waiting for GHA checks?

jrcrespoh:
> We might want to make a subset of https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/build.sh#L111 with the `RunXcodebuild()` function in the scripts directory of this repo as well.

@paulb777 Should I go ahead and look into this and incorporating it with abtesting?

abtesting/Shared/AppConfig.swift
jrcrespoh: This will be a "larger" change (refactor both iOS 15 and previous portions, replace RemoteConfigFetchStatus extension with one for RemoteConfigFetchAndActivateStatus, remove activate `changed` functionality), so I'll make it the last commit in case we would rather stick with separate calls 👍

abtesting/Shared/AppConfig.swift
```diff
@@ -57,6 +57,31 @@ class AppConfig: ObservableObject {
     }
   }
 
+  @available(iOS 15, *)
```
paulb777, line 60: I believe you also need to test for swift 5.5 since earlier versions of Swift won't be able to compile `async`.

jrcrespoh: Done ✅

abtesting/Shared/AppConfig.swift
jrcrespoh: Will be a separate PR

jrcrespoh:
Should updating the NotificationCenter observer to either Combine or async/await be a separate PR?

paulb777 (approved):
LGTM to merge here and follow up in separate PRs for the other tasks.

abtesting/Shared/AppConfig.swift
```diff
@@ -57,6 +57,33 @@ class AppConfig: ObservableObject {
...
+      let remoteConfig = RemoteConfig.remoteConfig()
+      do {
+        let status = try await remoteConfig.fetch(withExpirationDuration: 0)
+        print("Config fetch completed with status: \(status.debugDescription)")
+        do {
+          let changed = try await remoteConfig.activate()
+          let value = remoteConfig["color_scheme"].stringValue ?? "nil"
+          if changed {
+            print("Remote Config changed to: \(value)")
+            DispatchQueue.main.async {
```
peterfriese, line 72: Wouldn't it be better to use @MainActor to ensure `colorScheme` is updated on the main actor?

Using DispatchQueue.main.async in combination with the new async/await features feels a bit like an anti-pattern to me.

See https://peterfriese.dev/swiftui-concurrency-essentials-part1/#updating-published-properties

jrcrespoh: Great catch, thanks! I will include this change in a separate PR 👍

jrcrespoh: Actually, is there an alternative other than using `@MainActor`? Since we need the app to be backward-compatible with previous versions of iOS, I am not sure how to conditionally add `@MainActor`...

peterfriese: There are a couple of options:
1) You could apply it to a function that needs to run on the main actor (and the #ifdef that function)
2) Add it to the class (and #ifdef it there)
3) Call `MainAction.run` directly: https://www.avanderlee.com/swift/mainactor-dispatch-main-thread/#using-the-main-actor-directly

jrcrespoh: Awesome, in that case I can apply it to the already conditionally-compiled async function. Thanks for the pointers!

PR #1226, opened 2021-07-16
RTDB Star feature implementation

judyliu14:
- fetch "My Posts" reload after logout/login
- added starDictionary variable and didTapStarButton function
- TODO: initialize variables in PostCell

### DIFF
--- a/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
+++ b/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
@@ -32,8 +32,8 @@ struct Post: Identifiable {
     self.author = author
     self.title = title
     self.body = body
-    self.starCount = 0
-    self.starDictionary = [:]
+    starCount = 0
+    starDictionary = [:]
   }
 
   init?(id: String, dict: [String: Any]) {
@@ -49,7 +49,7 @@ struct Post: Identifiable {
     self.title = title
     self.body = body
     self.starCount = starCount
-    self.starDictionary = [:]
+    starDictionary = [:]
   }
 
   init() {
--- a/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
+++ b/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
@@ -24,6 +24,7 @@ struct Post: Identifiable {
   var title: String
   var body: String
   var starCount: Int
+  var starDictionary: [String: Bool]
 
   init(id: String, uid: String, author: String, title: String, body: String) {
     self.id = id
@@ -31,7 +32,8 @@ struct Post: Identifiable {
     self.author = author
     self.title = title
     self.body = body
-    starCount = 0
+    self.starCount = 0
+    self.starDictionary = [:]
   }
 
   init?(id: String, dict: [String: Any]) {
@@ -47,9 +49,33 @@ struct Post: Identifiable {
     self.title = title
     self.body = body
     self.starCount = starCount
+    self.starDictionary = [:]
   }
 
   init() {
     self.init(id: "", uid: "", author: "", title: "", body: "")
+  }
+
+  mutating func didTapStarButton(isStarred: Bool) {
+    let ref: DatabaseReference = {
+      Database.database().reference()
+    }()
+    if isStarred {
+      // add current user to the post.starDictionary
+      starDictionary[uid] = true
+      ref.child("/posts/\(id)/starDictionary").setValue(starDictionary)
+      ref.child("/user-posts/\(uid)/\(id)/starDictionary").setValue(starDictionary)
+      // increment starCount on current post
+      ref.child("/posts/\(id)/starCount").setValue(starCount + 1)
+      ref.child("/user-posts/\(uid)/\(id)/starCount").setValue(starCount + 1)
+    } else {
+      // remove current user from the post.starDictionary
+      starDictionary.removeValue(forKey: uid)
+      ref.child("/posts/\(id)/starDictionary").setValue(starDictionary)
+      ref.child("/user-posts/\(uid)/\(id)/starDictionary").setValue(starDictionary)
+      // decrement starCount on current post
+      ref.child("/posts/\(id)/starCount").setValue(starCount - 1)
+      ref.child("/user-posts/\(uid)/\(id)/starCount").setValue(starCount - 1)
+    }
   }
 }
--- a/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
+++ b/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
@@ -25,10 +25,7 @@ class PostViewModel: ObservableObject, Identifiable {
   @Published var body: String
   @Published var starCount: Int
   @Published var userIDsStarredBy: [String: Bool]
-
-  private lazy var ref: DatabaseReference = {
-    Database.database().reference()
-  }()
+  var postRef: DatabaseReference!
 
   init(id: String, uid: String, author: String, title: String, body: String) {
     self.id = id
@@ -36,8 +33,8 @@ class PostViewModel: ObservableObject, Identifiable {
     self.author = author
     self.title = title
     self.body = body
-    self.starCount = 0
-    self.userIDsStarredBy = [:]
+    starCount = 0
+    userIDsStarredBy = [:]
   }
 
   init?(id: String, dict: [String: Any]) {
@@ -45,6 +42,7 @@ class PostViewModel: ObservableObject, Identifiable {
     guard let author = dict["author"] as? String else { return nil }
     guard let title = dict["title"] as? String else { return nil }
     guard let body = dict["body"] as? String else { return nil }
+    let userIDsStarredBy = dict["userIDsStarredBy"] as? [String: Bool] ?? [:]
     let starCount = dict["starCount"] as? Int ?? 0
 
     self.id = id
@@ -53,24 +51,75 @@ class PostViewModel: ObservableObject, Identifiable {
     self.title = title
     self.body = body
     self.starCount = starCount
-    self.userIDsStarredBy = [:]
+    self.userIDsStarredBy = userIDsStarredBy
   }
 
-  func getCurrentUserID() -> String {
-    guard let currentUserID = Auth.auth().currentUser?.uid else { return "" }
-    return currentUserID
+  func isStarred() -> Bool {
+    if let uid = Auth.auth().currentUser?.uid {
+      print("first section ", userIDsStarredBy[uid] ?? false)
+      return userIDsStarredBy[uid] ?? false
+    }
+    print("second section false")
+    return false
   }
 
   func didTapStarButton() {
-
-    let currentUserID = self.getCurrentUserID()
-    let isStarred = userIDsStarredBy["\(currentUserID)"] ?? false
-    if isStarred {
-      // TODO: add current user to the post.starDictionary
-      userIDsStarredBy.removeValue(forKey: uid)
-    } else {
-      // TODO: remove current user from the post.starDictionary
-      userIDsStarredBy[uid] = true
+    // updating local values
+    if let uid = Auth.auth().currentUser?.uid {
+      if let _ = userIDsStarredBy[uid] {
+        starCount -= 1
+        userIDsStarredBy.removeValue(forKey: uid)
+      } else {
+        starCount += 1
+        userIDsStarredBy[uid] = true
+      }
     }
+
+    // updating firebase values
+    postRef = Database.database().reference().child("posts").child(id)
+    incrementStars(forRef: postRef)
+    postRef.observeSingleEvent(of: .value, with: { snapshot in
+      guard let value = snapshot.value as? [String: Any] else { return }
+      if let uid = value["uid"] as? String {
+        let userPostRef = Database.database().reference()
+          .child("user-posts")
+          .child(uid)
+          .child(self.id)
+        self.incrementStars(forRef: userPostRef)
+      }
+    })
+  }
+
+  func incrementStars(forRef ref: DatabaseReference) {
+    // [START post_stars_transaction]
+    ref.runTransactionBlock({ (currentData: MutableData) -> TransactionResult in
+      if var post = currentData.value as? [String: AnyObject],
+        let uid = Auth.auth().currentUser?.uid {
+        var userIDsStarredBy: [String: Bool]
+        userIDsStarredBy = post["userIDsStarredBy"] as? [String: Bool] ?? [:]
+        var starCount = post["starCount"] as? Int ?? 0
+        if let _ = userIDsStarredBy[uid] {
+          // Unstar the post and remove self from stars
+          starCount -= 1
+          userIDsStarredBy.removeValue(forKey: uid)
+        } else {
+          // Star the post and add self to stars
+          starCount += 1
+          userIDsStarredBy[uid] = true
+        }
+        post["starCount"] = starCount as AnyObject?
+        post["userIDsStarredBy"] = userIDsStarredBy as AnyObject?
+
+        // Set value and report transaction success
+        currentData.value = post
+        return TransactionResult.success(withValue: currentData)
+      }
+      return TransactionResult.success(withValue: currentData)
+    }) { error, committed, snapshot in
+      if let error = error {
+        print(error.localizedDescription)
+      }
+    }
+    // [END post_stars_transaction]
   }
 }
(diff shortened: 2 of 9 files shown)


database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/PostCell.swift
```diff
@@ -30,6 +38,7 @@ struct PostCellView: View {
             Image(systemName: isStarred ? "star.fill" : "star")
               .onTapGesture {
                 isStarred.toggle()
+                post.didTapStarButton(isStarred: isStarred)
```
maksymmalyhin, line 41: nit: Seems like we don't follow MVVM pattern here since Post is a clear Model and View should never modify Model directly but only via ViewModel. The intention of such indirection is needed to decouple data layer from the UI related logic which is important for creating maintainable applications. It is probably not a huge issue for our tiny sample app and such direct approach may be even an advantage for demo purposes, but I'm a little bit worried that we are showing a bad example.

If we decide to modify it to canonical MVVM I can help with options.

@paulb777 @ryanwilson @ncooke3 Do you have an opinion on if we should keep following canonical MVVM or should compromise on it for more direct code?

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
```diff
@@ -24,14 +24,16 @@ struct Post: Identifiable {
...
 
   init(id: String, uid: String, author: String, title: String, body: String) {
     self.id = id
     self.uid = uid
     self.author = author
     self.title = title
     self.body = body
-    starCount = 0
+    self.starCount = 0
+    self.starDictionary = [:]
```
maksymmalyhin, line 36: Ah, I see now. So `starDictionary` is a dictionary with user IDs that starred the post, right? If so, I think we should encode it in the name to improve readability e.g. `userIDsStarredBy` or similar.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
```diff
@@ -47,9 +49,33 @@ struct Post: Identifiable {
...
+
+  mutating func didTapStarButton(isStarred: Bool) {
+    let ref: DatabaseReference = {
+      Database.database().reference()
+    }()
+    if isStarred {
+      // add current user to the post.starDictionary
+      starDictionary[uid] = true
+      ref.child("/posts/\(id)/starDictionary").setValue(starDictionary)
+      ref.child("/user-posts/\(uid)/\(id)/starDictionary").setValue(starDictionary)
```
maksymmalyhin, line 67: nit: I cannot see the entire data schema but looks like we duplicate a post in two collections. I think we should rather have a collection of post IDs for a user and fetch the posts by ID from the post collection to avoid duplication.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
```diff
@@ -47,9 +49,33 @@ struct Post: Identifiable {
...
   }
+
+  mutating func didTapStarButton(isStarred: Bool) {
+    let ref: DatabaseReference = {
+      Database.database().reference()
+    }()
+    if isStarred {
+      // add current user to the post.starDictionary
+      starDictionary[uid] = true
+      ref.child("/posts/\(id)/starDictionary").setValue(starDictionary)
```
maksymmalyhin, line 66: nit: looks like we are writing the entire dictionary to modify add/remove a single user. In a real application this approach will lead to data loss, because of concurrent modification attempts from different users. E.g.:
1. user1 read the dict1
2.  user2 reads the dict1
3. user1 modifies and writes dict1+user1
4. user2 modifies and writes dict1+user2

As a result we have dict1+user2 stored, but should have dict1+user1+user2. This will only get worse with the number of users.

Suggestion to mitigate: add/remove a single child representing the user ID. It still may have issues with concurrent modification by the same user from different devices, but it is more an edge case. 

Though it's not a big issue for the sample project I still would recommend fixing it eventually to provide better guidance to Firebase users.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
maksymmalyhin: nit: An alternative could be maintaining a starred post IDs collection for a user. A downside: it will be more difficult to calculate starCount for a post.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
maksymmalyhin: BTW, the same is relevant for the starCount. You may count children in the star collection instead, but it may not perform well.  You can use [increment](https://firebase.google.com/docs/reference/swift/firebasedatabase/api/reference/Classes/ServerValue?authuser=0#increment_:) method but decrementing will be more challenging to implement correctly. TBH, taking into account complexity of correct and optimal counter implementation I would suggest excluding this feature so far.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/PostCell.swift
paulb777: My opinion is that we should strive to demonstrate best practices with our Quick Starts.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
judyliu14: Sounds good. I started the fetching and posting data structure by following the legacy example. Since SwiftUI has a slight different implementation, I will create a design doc that talks about my intended design structure for the app.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
judyliu14: Thank you for the feedback. I will take a closer look at updating the firebase data and use specific methods instead of updating the entire one.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
maksymmalyhin: Ah, I didn't know we'd had this data structure already. Then, I think, we can postpone modifying it and take it as a separate project and follow the existing approach so far.
Though to make the future refactoring easier we should isolate the Firebase Database logic and expose as little implementation details as possible to the rest of the app. I would suggest making `starDictionary` private and expose methods like `(un)startByUser(userID: String)` and `isStarredByUser(userID: String)`. In this case we will be able to change the implementation in the future without changing code that uses `Post` model.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/PostCell.swift
maksymmalyhin: The most straightforward way I can see so far would be the following:
1. Introduce a reference type (class) like `PostViewModel`. It will have all required `@Published` properties to display the post cell and the methods for actions like the star button press.
2. Replace `UserViewModel.posts` type by `[PostViewModel]`
3. Initialize `PostCell` with a selected `PostViewModel`. The `PostViewModel` will be a `@State` property then and will be the single source of truth

Using bindings will not be required then because `PostViewModel` is a reference type. In this case `PostViewModel` can depend on anything required like with Post, userID, Firebase instance, etc. and will provide a binding point between UI and model layers.

judyliu14: Thank you for the suggestion I think creating a `PostViewModel` is a good idea. I will make that change on the following commits.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
```diff
@@ -24,6 +24,7 @@ struct Post: Identifiable {
   var title: String
   var body: String
   var starCount: Int
+  var starDictionary: [String: Bool]
```
maksymmalyhin, line 27: I wonder if `Post` model is used anywhere now? In a real app we would have more rich Model layer where we would encapsulate Firebase logic. For demo purposes using RTDB directly from View Models looks fine to me unless we have to duplicate the logic between View Models. So it looks like `Post` model can be removed so far.

@paulb777 @ryanwilson @ncooke3 Any thoughts on it?

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+import Firebase
+
+class PostViewModel: ObservableObject, Identifiable {
+  @Published var id: String
```
maksymmalyhin, line 21: nit: View Model should expose only properties required for View. Is `id` displayed anywhere? The same question is relevant for the rest of the published properties.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+    self.id = id
+    self.uid = uid
+    self.author = author
+    self.title = title
+    self.body = body
+    self.starCount = starCount
+    self.userIDsStarredBy = userIDsStarredBy
+  }
+
+  func isStarred() -> Bool {
```
maksymmalyhin, line 57: nit: It can be defined as a computed property like:
```
var isStarred: Bool {
 return true
}
```

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+    if let uid = Auth.auth().currentUser?.uid {
+      return userIDsStarredBy[uid] ?? false
```
maksymmalyhin, line 59: nit: I think we can improve implementation because the current one has a couple issues:
1. If `Auth.auth().currentUser?.uid` updates so that `isStarred` value changes, the view will not be updated
2. property `userIDsStarredBy` is an implementation detail and should not be exposed to the view but currently it's the one that triggers view re-asking `isStarred` when needed.

It can be done in a separate PR.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+          .child("user-posts")
+          .child(uid)
+          .child(self.id)
+        self.incrementStars(forRef: userPostRef)
+      }
+    })
+  }
+
+  func incrementStars(forRef ref: DatabaseReference) {
+    // [START post_stars_transaction]
```
paulb777, line 92: These START and END tags are for connecting to code snippets from the docs. Have you coordinated with @morganchen12 about the implications?

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+  func isStarred() -> Bool {
+    if let uid = Auth.auth().currentUser?.uid {
+      return userIDsStarredBy[uid] ?? false
+    }
+    return false
+  }
+
+  func didTapStarButton() {
+    // updating local values
+    if let uid = Auth.auth().currentUser?.uid {
```
maksymmalyhin, line 66: optional: We use `Auth.auth().currentUser?.uid` multiple times. I think it can be extracted into a separate computed property.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+          .child(uid)
+          .child(self.id)
+        self.incrementStars(forRef: userPostRef)
+      }
+    })
+  }
+
+  func incrementStars(forRef ref: DatabaseReference) {
+    // [START post_stars_transaction]
+    ref.runTransactionBlock({ (currentData: MutableData) -> TransactionResult in
```
maksymmalyhin, line 93: Nice cleanup! Looks like this should solve most of the possible race conditions.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+      if var post = currentData.value as? [String: AnyObject],
+        let uid = Auth.auth().currentUser?.uid {
```
maksymmalyhin, line 95: From the [docs](https://firebase.google.com/docs/reference/swift/firebasedatabase/api/reference/Classes/DatabaseReference?authuser=0#runtransactionblock_:) it looks like the transaction block may be called several times. We should make sure that no state except `currentData` changes between the calls, so we should get `uid` outside of the transaction block.

maksymmalyhin (approved):
Mostly LGTM. Nice database logic cleanup! A couple nits, most of them can be addressed in following PRs

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/Post.swift
```diff
@@ -47,9 +49,33 @@ struct Post: Identifiable {
...
+      ref.child("/posts/\(id)/starCount").setValue(starCount + 1)
+      ref.child("/user-posts/\(uid)/\(id)/starCount").setValue(starCount + 1)
+    } else {
+      // remove current user from the post.starDictionary
+      starDictionary.removeValue(forKey: uid)
+      ref.child("/posts/\(id)/starDictionary").setValue(starDictionary)
+      ref.child("/user-posts/\(uid)/\(id)/starDictionary").setValue(starDictionary)
+      // decrement starCount on current post
+      ref.child("/posts/\(id)/starCount").setValue(starCount - 1)
+      ref.child("/user-posts/\(uid)/\(id)/starCount").setValue(starCount - 1)
```
morganchen12, line 78: If you're going to go down this route, you should use a [transaction](https://firebase.google.com/docs/database/ios/read-and-write#save_data_as_transactions).

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
morganchen12: No START or END tags are necessary in the new samples, please just assign any includecode bugs to me and I'll update the paths to point to the new locations of the old samples.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/UserViewModel.swift
```diff
@@ -127,43 +115,49 @@ class UserViewModel: ObservableObject {
...
-    // read data by listening for value events
-    refHandle = postRef.observe(DataEventType.value, with: { snapshot in
-      // retrieved data is of type dictionary of dictionary
-      guard let value = snapshot.value as? [String: [String: Any]] else { return }
-      // sort dictionary by keys (most to least recent)
-      let sortedValues = value.sorted(by: { $0.key > $1.key })
-      // store content of sorted dictionary into "posts" variable
-      self.posts = sortedValues.compactMap { Post(id: $0, dict: $1) }
-    })
+  func getPosts(tabOpened: String) {
```
morganchen12, line 136: Make `tabOpened` an enum with a raw value of String for better case-handling below.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/UserViewModel.swift
```diff
@@ -127,43 +115,49 @@ class UserViewModel: ObservableObject {
...
+          .child(userID)
+        fetchPosts(forRef: userPostListRef, tabOpened: tabOpened)
+      } else {
+        print("error: fetch myPosts was not successful")
+      }
+    }
   }
 
-  func fetchMyPosts() {
+  func fetchPosts(forRef ref: DatabaseReference, tabOpened: String) {
```
morganchen12, line 152: nit: Swift naming guidelines discourage duplicating type information in parameter labels. Consider `fetchPosts(for:tabOpened)` or `fetchPosts(from:tabOpened)`.

morganchen12 (commented):
Making great progress!

PR #1231, opened 2021-07-20
Implement My Top Posts View

judyliu14:
- added my top posts view
- fixed refresh error between different tabs that occurs when starring a post or creating a new post

### DIFF
--- a/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
+++ b/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
@@ -32,6 +32,7 @@ class PostViewModel: ObservableObject, Identifiable {
     }
     return false
   }
+  private var refHandle: DatabaseHandle?
 
   init(id: String, uid: String, author: String, title: String, body: String) {
     self.id = id
@@ -61,17 +62,6 @@ class PostViewModel: ObservableObject, Identifiable {
   }
 
   func didTapStarButton() {
-    // updating local values
-    if let uid = Auth.auth().currentUser?.uid {
-      if let _ = userIDsStarredBy[uid] {
-        starCount -= 1
-        userIDsStarredBy.removeValue(forKey: uid)
-      } else {
-        starCount += 1
-        userIDsStarredBy[uid] = true
-      }
-    }
-
     // updating firebase values
     postRef = Database.database().reference().child("posts").child(id)
     incrementStars(forRef: postRef)
@@ -84,6 +74,17 @@ class PostViewModel: ObservableObject, Identifiable {
           .child(self.id)
         self.incrementStars(forRef: userPostRef)
       }
+    })
+
+    // updating local values
+    updateStars(forRef: postRef)
+  }
+
+  func updateStars(forRef ref: DatabaseReference) {
+    refHandle = ref.observe(DataEventType.value, with: { snapshot in
+      guard let post = snapshot.value as? [String: AnyObject] else { return }
+      self.starCount = post["starCount"] as? Int ?? 0
+      self.userIDsStarredBy = post["userIDsStarredBy"] as? [String: Bool] ?? [:]
     })
   }
 
--- a/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
+++ b/database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
@@ -76,13 +76,11 @@ class PostViewModel: ObservableObject, Identifiable {
         self.incrementStars(forRef: userPostRef)
       }
     })
-
-    // updating local values
-    updateStars(forRef: postRef)
   }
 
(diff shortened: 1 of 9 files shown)


database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -33,6 +33,8 @@ class PostViewModel: ObservableObject, Identifiable {
     return false
   }
 
+  private var refHandle: DatabaseHandle?
```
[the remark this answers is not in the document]
judyliu14: On the new commit I have removed the handler with .onDisappear but am not sure if I have implemented correctly.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/MyPostsView.swift
```diff
@@ -31,6 +31,9 @@ struct MyPostsView: View {
       .onAppear {
         user.getPosts(tabOpened: tabOpened)
       }
+      .onDisappear {
+        user.removeRefHandles()
```
maksymmalyhin, line 35: Just a small correction. I don't think the view needs to know about ref handles as it is an implementation detail. The view should just forward the events important for the View Model. For this case it basically means renaming the method `UserViewModel.removeRefHandles()` to `UserViewModel.onViewDisappear()`.  The same applicable for `onAppear` here and for other views below.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/TopPostsView.swift
```diff
@@ -18,16 +18,36 @@ import SwiftUI
 import Firebase
 
 struct TopPostsView: View {
```
maksymmalyhin, line 20: How much different are `RecentPostsView` and `TopPostsView`? Can we have a generic `PostsView` that accepts a parameter like `type` that can be used instead of `tabOpened`?

maksymmalyhin (approved):
Mostly LGTM, a couple minor comments.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/TopPostsView.swift
judyliu14: Do you mean renaming tabOpened to type or something else? RecentPostView and TopPostsView are retrieving data from two different parts of Firebase Realtime Database.

maksymmalyhin: My main question is if `RecentPostsView` and `TopPostsView` display a given list of posts anyhow different? It looks to me that no and the only difference is `tabOpened` value. If it is so, I don't think we need to have two different classes, we can just have a single one and initialize it with `tabOpened` value.

My second point is about `tabOpened` name itself. It seems not exactly precise to me. If we have a view for a list of posts I would name the parameter like `postType` or something along these lines. I'm not too strong about it though.

judyliu14: I see did you mean combining all 3 `RecentPostsView`, `MyPostsView`, and `TopPostsView` into one file and pass in the tabOpened value? Currently in each view it contains its own `@StateObject` of type `UserViewModel` that refreshes on appear, it seems like the app is still functional after combining the 3 views into one file. So to clarify I will pass in the `tabOpened` and `navigation title` to a single class instead of having different classes?

maksymmalyhin: SGTM

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/UserViewModel.swift
```diff
@@ -138,10 +137,11 @@ class UserViewModel: ObservableObject {
   }
 
   func getPosts(tabOpened: String) {
-    if tabOpened == Tab.recentPosts.rawValue {
+    switch tabOpened {
```
paulb777, line 140: Can tabOpened be an enum instead of a String to avoid all the `rawValue`s?

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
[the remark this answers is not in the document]
morganchen12: Looks fine to me 👍

morganchen12 (approved):
LGTM, but please wait for all other approvers before merging.

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/TopPostsView.swift
judyliu14: Sounds good. Will restructure on the next PR.

paulb777 (approved):

PR #1238, opened 2021-07-30
Add SwiftUI performance quickstart

jrcrespoh:
TODO:
- [x] Fix CI
- [x] Remove macOS platform
- [x] Update platform versions
- [ ] Add upload functionality
- [ ] Add UI tests
- [ ] Add comments
- [ ] Consider adding Performance custom metrics
- [ ] Consider restructuring data flow
- [ ] Consider vision


performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -0,0 +1,217 @@
...
+
+      do {
+        let (data, response) = try await URLSession.shared.data(from: url)
+
+        guard let httpResponse = response as? HTTPURLResponse,
+          (200 ... 299).contains(httpResponse.statusCode) else {
+          print("Did not receive acceptable response: \(response)")
+          await updateStatusAsync(to: .failure)
+          return
+        }
```
morganchen12, line 71: nit: print the error if the http response isn't a 2xx code.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -0,0 +1,217 @@
...
+
+    @MainActor @available(iOS 15, tvOS 15, *) func updateActionAsync(to newAction: ProcessAction?) {
+      action = newAction
+    }
+
+    @available(iOS 15, tvOS 15, *) func downloadImageAsync() async {
+      await updateStatusAsync(to: .running)
+
+      guard let url = URL(string: site)
+      else {
```
morganchen12, line 57: nit: put the else on the same line as the guard

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -0,0 +1,217 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+import FirebasePerformance
+import Vision
+
+class Process: ObservableObject {
```
morganchen12, line 21: Maybe name this `LongRunningProcess`? Also, do you think it'd be a good idea to separate the class responsible for uploading the image from the class responsible for classifying it?

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -0,0 +1,217 @@
...
+  @Published var status: ProcessStatus = .idle
+  @Published var action: ProcessAction?
+  @Published var image: UIImage?
+  var categories: [(category: String, confidence: VNConfidence)] = []
+  let precision: Float = 0.7
+  let trace = Performance.sharedInstance().trace(name: "Classification")
```
morganchen12, line 27: This is only used in one function, so instead of creating it on initialization and retaining it you should create and use it in the function itself. This will prevent issues like accidentally reusing traces that have already been stopped (for example if you call `downloadImageAsync()` twice).

morganchen12 (commented):
Nice work! I left some feedback, but please wait for Maksym/Paul's review before merging.

.github/workflows/performance.yml
```diff
@@ -49,3 +49,65 @@ jobs:
...
+            (xcodebuild \
+              -workspace "$workspace_path" \
+              -scheme "${SAMPLE}Example${SWIFT_SUFFIX}" \
+              -sdk iphonesimulator \
+              -destination 'platform=iOS Simulator,name=iPhone 11' \
+              build \
+              | xcpretty) || EXIT_STATUS=$?
+            echo "Missing secrets: tests did not run."
+          fi
+          exit $EXIT_STATUS
```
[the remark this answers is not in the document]
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
```diff
@@ -16,9 +16,9 @@
...
-		5A81EBDA3528749141B3873C /* Pods_PerformanceExample__iOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = F4171CE0A02526CAA1C76922 /* Pods_PerformanceExample__iOS_.framework */; };
-		62EF508442D77FC8E18D3D05 /* Pods_PerformanceExample__tvOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 5881298C295F336EB049ADBE /* Pods_PerformanceExample__tvOS_.framework */; };
-		E0289F7ACA086958C3091522 /* Pods_Tests_iOS.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A7DBF8AF48A3164ECEA5CD0E /* Pods_Tests_iOS.framework */; };
+		84EF5C77BCF64159DC1554B3 /* Pods_Tests_iOS.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2766988506B7F4F19C97E851 /* Pods_Tests_iOS.framework */; };
+		92569034A655AF847A57DC68 /* Pods_PerformanceExample__tvOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = DA26F920681B148132EA1D2D /* Pods_PerformanceExample__tvOS_.framework */; };
+		E39F69BE4EC5DA0B8BCB2B3B /* Pods_PerformanceExample__iOS_.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 33530B44769A29407676BEA8 /* Pods_PerformanceExample__iOS_.framework */; };
```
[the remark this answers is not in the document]
jrcrespoh: Since I will be squashing and merging, any intermediary changes in these commits will not show up in the diff (since this is a new directory all files committed will be wholly new).

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/MainView.swift
```diff
@@ -0,0 +1,97 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+
+struct ImageView: View {
```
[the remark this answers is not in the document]
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/MainView.swift
```diff
@@ -0,0 +1,97 @@
...
+    VStack {
+      if let uiImage = process.image {
+        Image(uiImage: uiImage).padding(.bottom)
+        if process.action == .classify {
+          if process.categories.isEmpty {
+            Button("Classify Image") {
+              if #available(iOS 15, tvOS 15, *) {
+                #if swift(>=5.5)
+                  Task { await process.classifyImageAsync() }
+                #endif
```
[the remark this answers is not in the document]
jrcrespoh: My understanding is that this would make it necessary to use Xcode 13 even if compiling for iOS/tvOS 14.x, which is supported. Our CI wouldn't be able to compile the project either. I wanted to avoid the duplication with the else branch of the `#available` iOS/tvOS version check, but I can simply let the else branch of the `#if` swift version check be `process.classifyImage()` which I imagine would be less disruptive, so I'll take this approach unless you suggest otherwise.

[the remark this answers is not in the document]
morganchen12: Nah, let's just leave it as-is. Thanks for clarifying!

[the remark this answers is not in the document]
jrcrespoh: We would indeed have a problem with the current code if for some reason we are compiling for iOS/tvOS 15 but using Xcode 12 (which I don't think is possible?) or setting Xcode 13 to compile using lower versions of Swift (I thought it too unlikely), so I'll just add the extra else branch just in case.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: I'm printing the response since I don't believe I have access to an error in that scope. My understanding is that I only have access to an error if the `try await` function call throws an error.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: I didn't implement this because I felt the classification task was really just one "relatively simple" function, to which the `precision` user setting is passed. The `categories` is more associated with the image than with a `Classifier`. Additionally, this would require message passing between `Classifier` and `LongRunningProcess` to update the status of the ideally "sole running task". Doing this might make the app more complicated than it needs to be? I would appreciate hearing your thoughts on this, especially if you were thinking of some advantages to separating them.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
paulb777: `pod deintegrate` still needs to be run to avoid adding Pods references to the repo.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
```diff
@@ -0,0 +1,34 @@
+# Uncomment the next line to define a global platform for your project
+
+target 'PerformanceExample (iOS)' do
+  platform :ios, '14.0'
```
paulb777, line 4: Now that macOS is gone, can the contents of each target be factored outside to the top as one copy instead of three?

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
```diff
@@ -0,0 +1,34 @@
...
+  use_frameworks!
+
+  # Pods for PerformanceExample (iOS)
+  pod 'Firebase'
+  pod 'FirebasePerformance'
+
+end
+
+target 'Tests_iOS' do
+  platform :ios, '14.0'
```
paulb777, line 15: Is 13.0 feasible?

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
```diff
@@ -0,0 +1,34 @@
...
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
+
+  # Pods for PerformanceExample (iOS)
+  pod 'Firebase'
+  pod 'FirebasePerformance'
+
+end
+
+target 'Tests_iOS' do
```
paulb777, line 14: There should be Tests for tvOS as well

paulb777 (approved):
I'm ok to defer the comments except for the `pod deintegrate` to a future PR.

.github/workflows/performance.yml
```diff
@@ -49,3 +49,37 @@ jobs:
         env:
           SWIFT_SUFFIX: 'Swift'
           SAMPLE: Performance
+      - name: Build and Test SwiftUI iOS
+        run: |
+          set -io pipefail
+          cd performance/SwiftUIPerformanceQuickstart/PerformanceExample
+          bundle exec pod install --repo-update
+          xcodebuild \
```
granluo, line 57: The test is for PRs with secrets access only. We might only run build for forks.

.github/workflows/performance.yml
```diff
@@ -49,3 +49,37 @@ jobs:
...
+           -sdk iphonesimulator \
+           -destination 'platform=iOS Simulator,name=iPhone 11' \
+           build \
+           test \
+           | xcpretty
+        env:
+          SWIFT_SUFFIX: ' (iOS)'
+          SAMPLE: Performance
+          workspace_path: PerformanceExample.xcworkspace
+      - name: Build and Test SwiftUI tvOS
```
granluo, line 69: Could we use a [matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) here for ios and tvos

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
morganchen12: SGTM

.github/workflows/performance.yml
jrcrespoh: I'm not sure how to incorporate it since `scheme`, `sdk`, and `destination` only correspond with either iOS or tvOS and shouldn't be mixed and matched...

.github/workflows/performance.yml
jrcrespoh: Should I revert to checking for secrets?

granluo: As discussed in the meeting, we can revert it to the previous if-else version, but `set output` to let the `have_secrets` be public to other steps.

.github/workflows/performance.yml
granluo: Yep, I think to apply `matrix`, we might need a script to handle different params/flags. We can leave it like that before we have a script for all tests.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
jrcrespoh: Some SwiftUI features are not available and would break the app.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Podfile
jrcrespoh: Done ✅

.github/workflows/performance.yml
jrcrespoh: Because these scripts will later be consolidated, I instead used `check_secrets.sh` to achieve this.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,736 @@
...
+		4102055026A631F4004B24B3 = {
+			isa = PBXGroup;
+			children = (
+				419AADCD26B86A94005C4817 /* GoogleService-Info.plist */,
+				413CB73826B4AE8200E995E2 /* PerformanceExample--tvOS--Info.plist */,
+				4102055526A631F4004B24B3 /* Shared */,
+				4102057726A63210004B24B3 /* Tests_iOS */,
+				414C55A526B87EFB00114089 /* Tests_tvOS */,
+				4102055E26A631F5004B24B3 /* Products */,
+				CF3281DC37F16927E92A3A24 /* Pods */,
```
paulb777, line 97: Remove

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,736 @@
...
+		};
+		414C55A526B87EFB00114089 /* Tests_tvOS */ = {
+			isa = PBXGroup;
+			children = (
+				414C55A626B87EFB00114089 /* Tests_tvOS.swift */,
+			);
+			path = Tests_tvOS;
+			sourceTree = "<group>";
+		};
+		CF3281DC37F16927E92A3A24 /* Pods */ = {
```
paulb777, line 139: Remove.

paulb777 (approved):
The lingering `Pods` references are the result of running `pod deintegrate` with the project open in Xcode.

It can be fixed by editing manually or by opening the xcodeproj and deleting the `Pods` in the Project Navigator. It should be red.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  func uploadImage() {}
+
+  func makeTrace(called name: String = "Classification") -> Trace? {
+    #if os(tvOS)
+      let platform = "tvOS"
+    #elseif os(iOS)
+      let platform = "iOS"
+    #endif
```
paulb777, line 187: Consider adding an `#else`  with a fatal_error.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/PerformanceExample.xcodeproj/project.pbxproj
jrcrespoh: Done ✅

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: Done ✅

paulb777 (approved):
💯

PR #1241, opened 2021-08-04
Database UI tests: fix removing all messages

maksymmalyhin:


ncooke3 (approved):
Thanks Maksym 👍

maksymmalyhin:
> Thanks! Do you know what changed to make this start failing?

I think, at some point during testing enough posts created under the testing account to trigger removal failure due to off-screen cells representing the posts. Once it starting failing, each test run created even more posts that were not removed making failure more possible. The updated approach should work for any amount of posts so the failure should not repeat. But let's test this assumption 😄.

PR #1243, opened 2021-08-05
FIAM: explicitly import modules used in Swift

maksymmalyhin:
Should fix the [failure](https://github.com/firebase/firebase-ios-sdk/runs/3251416541?check_suite_focus=true) from https://github.com/firebase/firebase-ios-sdk/issues/8471#issuecomment-893449537.

### DIFF
--- a/inappmessaging/InAppMessagingExampleSwift/ViewController.swift
+++ b/inappmessaging/InAppMessagingExampleSwift/ViewController.swift
@@ -17,7 +17,6 @@
 import UIKit
 
 import Firebase
-import FirebaseAnalytics
 
 class ViewController: UIViewController {
   @IBOutlet var textField: UITextField!
--- a/inappmessaging/InAppMessagingExampleSwift/ViewController.swift
+++ b/inappmessaging/InAppMessagingExampleSwift/ViewController.swift
@@ -14,7 +14,10 @@
 //  limitations under the License.
 //
 
+import UIKit
+
 import Firebase
+import FirebaseAnalytics
 
 class ViewController: UIViewController {
   @IBOutlet var textField: UITextField!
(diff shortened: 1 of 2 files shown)


ncooke3 (approved):
LGTM. I wonder why these missing imports in FIAM qs and SDK didn't cause failures previously...

inappmessaging/InAppMessagingExampleSwift/ViewController.swift
```diff
@@ -14,7 +14,10 @@
 //  limitations under the License.
 //
 
+import UIKit
+
 import Firebase
+import FirebaseAnalytics
```
paulb777, line 20: This one shouldn't be necessary because FirebaseAnalytics should come with the Firebase import.

paulb777 (approved):
Xcode can be random about auto-importing UIKit.

maksymmalyhin:
> LGTM. I wonder why these missing imports in FIAM qs and SDK didn't cause failures previously...

Good question. I don't know for sure, I guess UIKit was imported implicitly in some way before. But importing all required modules in a Swift file is a best practice so it seems like a reasonable change to me.

inappmessaging/InAppMessagingExampleSwift/ViewController.swift
maksymmalyhin: True... Let me take a closer look. I have an error locally when FirebaseAnalytics is not imported:
Screen Shot 2021-08-05 at 10 22 09 AM

maksymmalyhin: It seems to be related to flakes with Xcode module cache or something. I guess I had in a broken state. After removing DerivedData `Analytics` was marked as missing before the first build. I removed the import.

PR #1246, opened 2021-08-11
Add UI tests for performance [SwiftUI]

jrcrespoh:


performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
```diff
@@ -217,5 +217,5 @@ enum ProcessStatus {
 enum ProcessAction: String, CaseIterable {
   case download = "Download"
   case classify = "Classify"
-  case upload = "Upload"
+//  case upload = "Upload"
```
paulb777, line 220: Why is this commented?

paulb777 (approved):
Thanks!

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/Process.swift
jrcrespoh: The upload functionality will be added as a follow-up PR, and only download and classify are tested with the UI tests.

morganchen12: maybe add a TODO comment?

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/UITests.swift
```diff
@@ -0,0 +1,150 @@
...
+    let element = app.staticTexts[text]
+    XCTAssert(element.waitForExistence(timeout: timeout), "Missing text '\(text)'.")
+    XCTAssert(element.isHittable, "Text '\(text)' not in view.")
+  }
+
+  func checkStatus(_ status: String, timeout: TimeInterval = 1) throws {
+    #if os(iOS)
+      try checkText(status, timeout: timeout)
+    #endif
+  }
```
morganchen12, line 55: nit: you can wrap this whole function in the `#if` so it doesn't show up in the sidebar for non-iOS test runs.

morganchen12 (approved):
LGTM, had some small comments.

performance/SwiftUIPerformanceQuickstart/PerformanceExample/Shared/UITests.swift
jrcrespoh: I'm not sure which sidebar you mean, but if you mean the "Test navigator" with the tests that run then it doesn't even show up since it is not an independent test (because the function name doesn't start with `test`):

image

Also, if I were to do this then I would have to do the same `#if` wrapping every time I call it within `testAllViews`, which is definitely not optimal. Please let me know if you think we should look for a different approach and not keep it as is.

morganchen12: rationale LGTM!

PR #1248, opened 2021-08-12
Add region tags

rosalyntan:
Add a couple of region tags for the Firebase public docs to reference.

### DIFF
--- a/.github/workflows/crashlytics.yml
+++ b/.github/workflows/crashlytics.yml
@@ -51,7 +51,7 @@ jobs:
           SWIFT_SUFFIX: 'Swift'
           SAMPLE: Crashlytics
           LEGACY: true
-      - name: Build SwiftUI
+      - name: Build SwiftUI (iOS)
         run: |
           cd crashlytics
           ../scripts/install_prereqs/crashlytics.sh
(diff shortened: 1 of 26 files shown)


rosalyntan:
Hmm `check / check (pull_request)` is failing with

```
Run scripts/style.sh test-only
authentication/LegacyAuthQuickstart/AuthenticationExampleSwift/MainViewController.swift needs formatting.
Proposed commit is not style compliant.
Run scripts/style.sh and git add the result.
Error: Process completed with exit code 1.
```

However, when I run `scripts/style.sh` nothing happens. I don't see anything obviously styled incorrectly in the `MainViewController.swift` either.

paulb777:
@rosalyntan Perhaps turn on `set -x` to see what is happening when you run style?  And/or merge #1249.

Any reason for adding the new tags to the Legacy quickstart?  Shouldn't the primary quickstart be updated with them?

rosalyntan:
Thanks! Merged #1249 into this PR.

> Any reason for adding the new tags to the Legacy quickstart? Shouldn't the primary quickstart be updated with them

I added the region tags to the legacy quickstart because all the existing auth references in the snippet registry reference the legacy quickstart: https://source.corp.google.com/piper///depot/google3/googledata/devsite/_common/en/snippet_registry.yaml;l=2653-2873

morganchen12 (approved):
Do the snippets in google3 need to be migrated?

rosalyntan:
> Do the snippets in google3 need to be migrated?

Not sure, I figured the auth snippets were using the legacy quickstart for a reason (and it looks like analytics and remote config snippets do the same).

PR #1272, opened 2021-08-27
Add UI Tests for Crashlytics

judyliu14:
- [x] iOS
- [x] tvOS
- [x] macOS
- [x] WatchOS: Running watchOS would use iOS test

### DIFF
--- a/crashlytics/Shared/UITests.swift
+++ b/crashlytics/Shared/UITests.swift
@@ -35,18 +35,6 @@ class UITests: XCTestCase {
     XCTAssertTrue(app.buttons["Crash"].isHittable, "Crash button is missing from the view.")
   }
 
-  func testDynamicUI() throws {
-    let app = XCUIApplication()
-    app.launch()
-    #if os(iOS) || os(watchOS)
-      app.buttons["Crash"].tap()
-    #elseif os(macOS)
-      app.buttons["Crash"].click()
-    #elseif os(tvOS)
-      XCUIRemote.shared.press(.select)
-    #endif
-  }
-
   func testLaunchPerformance() throws {
     if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 7.0, *) {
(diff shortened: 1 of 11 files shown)


crashlytics/Shared/UITests.swift
```diff
@@ -0,0 +1,72 @@
...
+    app.launch()
+
+    // Use recording to get started writing UI tests.
+    // Use XCTAssert and related functions to verify your tests produce the correct results.
+
+    #if !os(macOS)
+      XCTAssertTrue(
+        app.navigationBars["Crashlytics Example"].exists,
+        "Crashlytics Example is missing from the navigation bar"
+      )
```
morganchen12, line 45: Assertions should be written affirming the point you expect to be true, because if the test fails they'll show up in the logs as
```
expected "Crashlytics Example exists in the navigation bar" to be true
```

crashlytics/Shared/UITests.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    // In UI tests it is usually best to stop immediately when a failure occurs.
+    continueAfterFailure = false
+
+    // In UI tests it’s important to set the initial state - such as interface orientation - required for your tests before they run. The setUp method is a good place to do this.
+  }
+
+  override func tearDownWithError() throws {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+  }
```
morganchen12, line 31: nit: The auto-generated comments in the setup and teardown functions can be removed.

morganchen12 (approved):

crashlytics/Shared/UITests.swift
judyliu14: The second string of XCTAssertTrue is a description of the failure. https://developer.apple.com/documentation/xctest/xctasserttrue

morganchen12: Whoops, thanks for the correction :)

PR #1276, opened 2021-09-01
Add async/await to rtdb

judyliu14:


database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Models/PostViewModel.swift
```diff
@@ -101,6 +101,24 @@ class PostViewModel: ObservableObject, Identifiable {
     })
   }
 
+  #if swift(>=5.5)
+    @available(iOS 15, macOS 12, tvOS 15, *)
```
paulb777, line 105: @available(iOS 15, tvOS 15, macOS 12, watchOS 8, *)

database/DatabaseExampleSwiftUI/DatabaseExample/Shared/Views/PostCell.swift
```diff
@@ -32,7 +32,17 @@ struct PostCell: View {
             #if os(iOS) || os(macOS)
               starButton
                 .onTapGesture {
-                  post.didTapStarButton()
+                  if #available(iOS 15, macOS 12, *) {
+                    #if swift(>=5.5)
+                      Task { await
```
paulb777, line 37: Why is a `Task` needed here? Can it be just `await post.didTapStarButtonAsync()`?

judyliu14: Since the code is executed within`.onTapGesture`, which does not have an async signature, we need to use Task to open the async gesture. So if I remove the `Task` it will result in an error message: `Cannot pass function of type '() async -> Void' to parameter expecting synchronous function type`

paulb777 (approved):
Thanks! LGTM on green

PR #1278, opened 2021-09-01
Fix problematic naming performance

jrcrespoh:


jrcrespoh:
Fixing https://github.com/firebase/firebase-ios-sdk/runs/3483398120?check_suite_focus=true

ncooke3 (approved):
thanks!

PR #1287, opened 2021-09-02
Add Async/Await to Crashlytics

judyliu14:
- add async await
- cleanup init() function with helper functions

### DIFF
--- a/crashlytics/Shared/ContentView.swift
+++ b/crashlytics/Shared/ContentView.swift
@@ -20,14 +20,14 @@ import Firebase
 struct ContentView: View {
   private var crashlyticsReference = Crashlytics.crashlytics()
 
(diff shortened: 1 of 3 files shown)


paulb777 (approved):
Thanks!

crashlytics/Shared/ContentView.swift
```diff
@@ -25,6 +45,17 @@ struct ContentView: View {
         Text("Crash")
       }
       .navigationTitle("Crashlytics Example")
+      .onAppear {
```
peterfriese, line 48: Have you considered using the `task` modifier?

judyliu14: Yes using the `.task` modifier is a good suggestion. I will make the change in a following PR.

PR #1288, opened 2021-09-03
Crashlytics Async/Await Update

judyliu14:
- update .onAppear modifier to .task modifier for (iOS 15, tvOS 15, macOS 12, watchOS 8, *)
- ensure fallback on earlier versions execute correctly

### DIFF
--- a/crashlytics/Shared/ContentView.swift
+++ b/crashlytics/Shared/ContentView.swift
@@ -17,6 +17,19 @@
 import SwiftUI
 import Firebase
 
+struct CrashButtonView: View {
+  var body: some View {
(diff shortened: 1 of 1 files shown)


paulb777 (approved):
Nice improvement. Thanks @peterfriese and @judyliu14

peterfriese:
> Nice improvement. Thanks @peterfriese and @judyliu14

This is a good example for what we discussed yesterday: once iOS 15 / Swift 5.5. go GA, the code can be simplified quite a bit :-)

peterfriese (approved):
This looks great, thanks for the update!

PR #1295, opened 2021-09-13
Rotate secrets

ncooke3:
Corresponding Firebase PR: https://github.com/firebase/firebase-ios-sdk/pull/8632 (cl/395792083)

PR #1296, opened 2021-09-14
Decrypted and then encrypted secrets tar ball with no changes

ncooke3:


ncooke3:
Well, CI confirms I don't know how to make a tarball.

ncooke3:
Ahh, there is a path issue when unzipping the tar file. see #1295

PR #1298, opened 2021-09-16
Use common qs bundle ID format in Crashlytics qs

ncooke3:
Was using crashlytics qs to debug an issue and got a warning that the bundle ID didnt match with that in the default googleservice-info.plist. 

I did this with: 
```
grep -rl 'Google-LLC.CrashlyticsSwiftUIExample' . | xargs sed -i '' 's/Google-LLC.CrashlyticsSwiftUIExample/com.google.firebase.quickstart.CrashlyticsExample/'
```


ncooke3:
Hmm im unsure why the [failure](https://github.com/firebase/quickstart-ios/pull/1298/checks?check_run_id=3626077004#step:4:197) is occurring. It's building locally on Xcode 13 and that view modifier is still called [`task`](https://developer.apple.com/documentation/swiftui/circle/task(priority:_:))

morganchen12 (approved):

PR #1313, opened 2021-10-16
Facebook 12.0.1 requires Xcode 13

paulb777:
See also facebook/facebook-ios-sdk#1911 and https://github.com/firebase/firebase-ios-sdk/pull/8814

### DIFF
--- a/.github/workflows/authentication.yml
+++ b/.github/workflows/authentication.yml
@@ -35,6 +35,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@master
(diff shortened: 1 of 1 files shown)


jawwad:
Couldn't help but notice how appropriate the issue number is for this issue, i.e. #1313 for requiring Xcode 13 :)

paulb777:
This PR is no longer immediately necessary now that Facebook has restored Xcode 12 with its 12.0.2 release. However, this PR will still make us more resilient for a future change.

PR #1328, opened 2021-12-22
Stop using Firebase umbrella pod in Storage quickstart

paulb777:
More context at https://github.com/firebase/firebase-ios-sdk/issues/9074

The rest of this repo should be similarly updated.


ncooke3 (approved):
thanks!

PR #1329, opened 2021-12-25
Messaging use async for iOS13

ykws:
If Firebase Messaging drop iOS12 or lower, recommend to use async.

https://github.com/firebase/firebase-ios-sdk/blob/e9d483222df04f0b3b41bf0f61c30d67e15f387e/FirebaseMessaging.podspec#L23


ykws:
Oh..

```swift
concurrency is only available in iOS 15.0.0 or newer

                   didReceiveRemoteNotification userInfo: [AnyHashable: Any]) async -> UIBackgroundFetchResult {
```

paulb777:
Thanks @ykws.  With Xcode 13.2, Swift Concurrency supports down to iOS 13.  However, GitHub Actions is not yet using 13.2 and many developers may not have updated yet.   We'll likely be able to merge this in a few months.

paulb777:
Merged in #1419.

Thanks for the contribution @ykws! 🎆

PR #1332, opened 2022-01-07
Update RC JSON example to use new API

paulb777:
Quickstart PR that accompanies https://github.com/firebase/firebase-ios-sdk/pull/9084 and the new type safe Swift Codable APIs for Remote Config

Now that 8.12.0 has released, this is functional with published podspecs.

### DIFF
--- a/config/ConfigExample/RemoteConfigViewController.swift
+++ b/config/ConfigExample/RemoteConfigViewController.swift
@@ -94,9 +94,9 @@ class RemoteConfigViewController: UIViewController {
 
   /// This method applies our remote config values to our UI
   private func updateUI() {
-    remoteConfigView.topLabel.text = remoteConfig["topLabelKey"].stringValue
+    remoteConfigView.topLabel.text = remoteConfig[stringValue: "topLabelKey"]
     updateJSONView()
-    remoteConfigView.bottomLabel.text = remoteConfig["bottomLabelKey"].stringValue
+    remoteConfigView.bottomLabel.text = remoteConfig[stringValue: "bottomLabelKey"]
   }
 
   // MARK: - Private Helpers
@@ -128,7 +128,7 @@ class RemoteConfigViewController: UIViewController {
       }
     }
 
-    struct Recipe: Decodable {
+    struct Recipe: Decodable, CustomStringConvertible {
       var recipe_name: String
       var ingredients: [String]
       var prep_time: Int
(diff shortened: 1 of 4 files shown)


config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -48,41 +49,41 @@ class RemoteConfigViewController: UIViewController {
...
+  }
+
   /// Initializes defaults from `RemoteConfigDefaults.plist` and sets config's settings to developer mode
   private func setupRemoteConfig() {
     remoteConfig = RemoteConfig.remoteConfig()
-    remoteConfig.setDefaults(fromPlist: "RemoteConfigDefaults")
+    // This is an alternative to remoteConfig.setDefaults(fromPlist: "RemoteConfigDefaults"))
+    try? remoteConfig.setDefaults(from: QSConfig(topLabelKey: "myTopLabel",
+                                                 bottomLabelKey: "Buy one get one free!",
+                                                 freeCount: 4))
```
morganchen12, line 65: nit: handle the returned error instead of ignoring it

config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -48,41 +49,41 @@ class RemoteConfigViewController: UIViewController {
...
-      }
-    }
-  }
-
   /// Fetches and activates remote config values
   @objc
   private func fetchAndActivateRemoteConfig() {
-    remoteConfig.fetchAndActivate { status, error in
-      guard error == nil else { return self.displayError(error) }
+    Task.init {
```
morganchen12, line 75: Does `Task {` not work here?

config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -48,41 +49,41 @@ class RemoteConfigViewController: UIViewController {
...
+      guard let _ = try? await remoteConfig.fetchAndActivate() else {
+        print("Failed to fetchAndActivate.")
+        return
+      }
```
morganchen12, line 79: nit: catch the error and use `displayError` to maintain parity with the old code

config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -48,41 +49,41 @@ class RemoteConfigViewController: UIViewController {
...
       print("Remote config successfully fetched & activated!")
+      do {
+        let qsConfig: QSConfig = try remoteConfig.decoded()
+        print(qsConfig)
+      } catch {
+        print("Failed to decode config.")
```
morganchen12, line 85: call `displayError` here as well

morganchen12 (approved):
LGTM with a few comments.

config/ConfigExample/RemoteConfigViewController.swift
paulb777: Done.

config/ConfigExample/RemoteConfigViewController.swift
paulb777: Good catch!

paulb777:
@morganchen12 Thanks for the review. Any idea why the UI Tests are crashing in CI?

paulb777:
Reproducing locally from command line - but not Xcode UI:

2022-02-11 15:01:03.919719-0900 ConfigExample[21355:6419222] [Animation] +[UIView setAnimationsEnabled:] being called from a background thread. Performing any operation from a background thread on UIView or a subclass is not supported and may result in unexpected and insidious behavior.

paulb777:
Something about the swift to the async/await fetchAndActivate API is triggering the issue. Reverting back to the old one for now.

config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -196,7 +209,7 @@ extension UIViewController {
   public func displayError(_ error: Error?, from function: StaticString = #function) {
```
ncooke3, line 209: I think the error you were running into had to do with how this method was being called within the async contexts. It was then being called on a background thread causing the crash. Weird how it wasn't straightforward to reproduce though. I think it would work if the implementation of this method was wrapped in `DispatchQueue.main.async { ... }`. That way, you could call it from whatever context and it would put itself on the main queue by itself.

paulb777: Thanks Nick!  If you have the bandwidth, would you see if you can get the async/await variation to work?  Otherwise I should be able to get it to work later this week.

PR #1351, opened 2022-02-16
Update Firestore Podfile.lock to latest

paulb777:
#1099 workaround


paulb777:
Note the firebase-ios-sdk zip tests will continue to be stuck on the 8.2 version of FirebaseFirestoreSwift because the zip tests override the minimum version to 10.

PR #1357, opened 2022-03-03
Update quickstart for non-optional Codable

morganchen12:
See https://github.com/firebase/firebase-ios-sdk/pull/9101, internal b/222563260.

This change will cause build failures before Firebase dependencies are updated to 8.13.0.


paulb777 (approved):
Thanks! Please update the Podfile and Podfile.lock as well and merge after 8.13.0 publishes.

PR #1365, opened 2022-04-01
Storage macOS Quickstart SwiftUI update

granluo:


storage/StorageExample/Shared/ContentView.swift
```diff
@@ -19,6 +19,10 @@ import FirebaseAuth
 import FirebaseStorage
 import FirebaseStorageSwift
 
+#if os(macOS)
+  typealias UIImage = NSImage
+  typealias UIApplication = NSWorkspace
```
morganchen12, line 24: I'd recommend not doing this since it's clever, but not extensible: If the app ever depends on an iOS or macOS specific image feature, the typealias will break. Instead, you could wrap this in an `Image` helper struct that references the correct platform-specific type below and then redefine methods on top of it.

storage/StorageExample/Shared/ContentView.swift
```diff
@@ -229,13 +271,25 @@ struct OrangeButton: ButtonStyle {
...
+    #if os(iOS)
+      jpegData(compressionQuality: 1)
+    #elseif os(macOS)
+      let cgImage = self.cgImage(forProposedRect: nil, context: nil, hints: nil)!
+      let bitmapRep = NSBitmapImageRep(cgImage: cgImage)
+      let jpegData = bitmapRep.representation(
+        using: NSBitmapImageRep.FileType.jpeg,
+        properties: [:]
+      )!
+      return jpegData
```
morganchen12, line 284: Instead of force-unwrapping in this method, you can just return `nil` since the return type is optional.

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,191 @@
...
+              continuation.resume(throwing: error)
+            } else {
+              do {
+                if FileManager.default.fileExists(atPath: destURL.path) {
+                  try FileManager.default.removeItem(at: destURL)
+                }
+                try FileManager.default.copyItem(at: url!, to: destURL)
+                continuation.resume()
+              } catch {
+                print("Cannot copy item at \(url!) to \(destURL): \(error)")
```
morganchen12, line 108: This should be passed to the continuation.

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,191 @@
...
+        return try await withCheckedThrowingContinuation { continuation in
+          self.loadObject(ofClass: UIImage.self) { image, error in
+            if let error = error {
+              continuation.resume(throwing: error)
+            }
+            continuation.resume(returning: image as? UIImage)
+          }
+        }
+      } catch {
+        print("Image was not properly loaded.")
```
morganchen12, line 129: This error should be rethrown.

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,191 @@
...
               }
-              try FileManager.default.copyItem(at: url!, to: destURL)
-              continuation.resume()
-            } catch {
-              print("Cannot copy item at \(url!) to \(destURL): \(error)")
             }
           }
         }
+      } catch {
+        print("Cannot load file from the image picker.")
```
morganchen12, line 114: This should be rethrown (you can likely just omit the top-level do/catch).

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,191 @@
...
+#elseif os(macOS)
+
+  struct ImagePicker: View {
+    @Binding var image: NSImage?
+    var imageURL: URL?
+
+    var body: some View {
+      ZStack {
+        if self.image != nil {
+          Image(nsImage: self.image!)
```
morganchen12, line 144: Use `if let image = self.image` to avoid the force-unwrap.

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,191 @@
...
+      .frame(height: 320)
+      .background(Color.black.opacity(0.6))
+      .cornerRadius(16)
+      .onDrop(of: ["public.url", "public.file-url"], isTargeted: nil) { (items) -> Bool in
+        if let item = items.first {
+          if let identifier = item.registeredTypeIdentifiers.first {
+            if identifier == "public.url" || identifier == "public.file-url" {
+              item.loadItem(forTypeIdentifier: identifier, options: nil) { urlData, error in
+                if let data = urlData as? Data {
+                  let urll = NSURL(
```
morganchen12, line 170: nit: `url` instead of `urll`

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -126,7 +127,7 @@ import SwiftUI
           }
         }
       } catch {
-        print("Image was not properly loaded.")
+        print("Image was not properly loaded.\(error)")
```
morganchen12, line 130: `throw` this error (here and throughout) instead of returning nil.

morganchen12 (commented):
LGTM with one last comment. Please wait for @ncooke3's approval before merging.

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -112,7 +117,8 @@ import SwiftUI
           }
         }
       } catch {
-        print("Cannot load file from the image picker.\(error)")
+        print("Cannot load file from the image picker.")
+        throw LoadingItemError.loadingTempFileError
```
morganchen12, line 121: Can you not just `throw error` instead of defining a separate error enum?

morganchen12 (approved):
LGTM

storage/StorageExample/Shared/ContentView.swift
```diff
@@ -72,15 +76,34 @@ struct ContentView: View {
...
+            Button {
+              NSWorkspace.shared.selectFile(
+                viewModel.fileLocalDownloadURL!.path,
+                inFileViewerRootedAtPath: ""
+              )
+            } label: {
+              Text("Open in Finder")
+            }
+            .buttonStyle(OrangeButton())
+
```
ncooke3, line 105: nit: remove new line 
```suggestion
```

storage/StorageExample/Shared/ContentView.swift
```diff
@@ -208,12 +239,34 @@ struct ContentView: View {
...
+  #elseif os(macOS)
+    func setImage(fromImage image: NSImage) -> Image {
+      return Image(nsImage: image)
+    }
+
+    func setImage(fromURL url: URL) -> Image {
+      return setImage(fromImage: NSImage(contentsOfFile: url.path)!)
+    }
+  #endif
```
ncooke3, line 269: optional: Does this work? If so, it avoids the need for duplicate `setImage(fromURL url: URL) -> Image` APIs.

```suggestion
  #if os(iOS)
    func setImage(fromImage image: UIImage) -> Image {
      return Image(uiImage: image)
    }
  #elseif os(macOS)
    func setImage(fromImage image: NSImage) -> Image {
      return Image(nsImage: image)
    }
  #endif

  func setImage(fromURL url: URL) -> Image {
    return setImage(fromImage: .init(contentsOfFile: url.path)!)
  }

```

storage/StorageExample/Shared/ContentView.swift
```diff
@@ -228,14 +281,31 @@ struct OrangeButton: ButtonStyle {
...
+    var jpeg: Data? {
+      let cgImage = self.cgImage(forProposedRect: nil, context: nil, hints: nil)!
+      let bitmapRep = NSBitmapImageRep(cgImage: cgImage)
+      return bitmapRep.representation(
+        using: NSBitmapImageRep.FileType.jpeg,
+        properties: [:]
+      )
+    }
+  }
+
```
ncooke3, line 302: nit: remove newline
```suggestion
```

storage/StorageExample/Shared/ViewModel.swift
```diff
@@ -15,20 +15,29 @@
...
+@MainActor
 class ViewModel: ObservableObject {
   @Published var image: Image?
   @Published var showingImagePicker = false
-  @Published var inputImage: UIImage?
+
+  #if os(iOS)
+    @Published var inputImage: UIImage?
+  #elseif os(macOS)
+    @Published var inputImage: NSImage?
```
ncooke3, line 31: Do you need to `import AppKit` if `os(macOS)`?

storage/StorageExample/Shared/ImagePicker.swift
```diff
@@ -13,119 +13,193 @@
...
-      print("Image was not properly loaded.")
-      return nil
+      .frame(height: 320)
+      .background(Color.black.opacity(0.6))
+      .cornerRadius(16)
+      .onDrop(of: ["public.url", "public.file-url"], isTargeted: nil) { (items) -> Bool in
+        if let item = items.first {
+          if let identifier = item.registeredTypeIdentifiers.first {
+            if identifier == "public.url" || identifier == "public.file-url" {
+              item.loadItem(forTypeIdentifier: identifier, options: nil) { urlData, error in
```
ncooke3, line 170: Curious, does an async/await API generate for this?
```swift
try await item.loadItem( //...
```

ncooke3:
Thanks @granluo, looks great!

storage/StorageExample/Shared/ImagePicker.swift
granluo: It looks like [loadItem](https://developer.apple.com/documentation/foundation/nsitemprovider/1403900-loaditem) does get an async note in its doc. Will update it.

storage/StorageExample/Shared/ViewModel.swift
granluo: I think `SwiftUI` is sufficient to handle the UI. Without importing `AppKit`, the app can be built and run successfully.

storage/StorageExample/Shared/ContentView.swift
granluo: That's neat! It is working :)

ncooke3: oh, cool!

ncooke3 (approved):
LGTM. nice work!

PR #1363, opened 2022-03-24
Update Storage iOS Sample.

granluo:

PR #1367, opened 2022-04-02
Remove GoogleService-Info.plist from test targets

paulb777:
Might fix https://github.com/firebase/firebase-ios-sdk/issues/9534 introduced by #1363


paulb777:
I'm going to force the merge and restart the failed firebase-ios-sdk nightlies.

paulb777:
Based on retry failure https://github.com/firebase/firebase-ios-sdk/runs/5800286013?check_suite_focus=true, this problem seems innocuous and unrelated.

granluo:
> Based on retry failure https://github.com/firebase/firebase-ios-sdk/runs/5800286013?check_suite_focus=true, this problem seems innocuous and unrelated.

`GoogleService-Info.plist` was in the target and the [script](https://github.com/firebase/quickstart-ios/blob/master/scripts/info_script.rb#L33) add another plist and so it causes the `duplicated tasks`. 

`GoogleService-Info.plist` has target memberships for al targets, and it is under the `Copy Bundle Resources` of `Build Phases`, and the [script](https://github.com/firebase/quickstart-ios/blob/master/scripts/info_script.rb#L33) above adds the `GoogleService-Info.plist` to the `Compile Sources` of `Build Phases`. And this cause the issue during build. I am not sure why other Quickstarts do not have such issue. 

Instead of adding the reference to the project, we can replace the existing plist with the expected one, since the project has the reference for targets already.

granluo:
`scripts/decrypt_gha_secret.sh` replaced the plist and the [PR](https://github.com/firebase/firebase-ios-sdk/runs/5804811164?check_suite_focus=true) can work for storage. Maybe we can test it for other Quickstarts? IIRC, all Quickstarts now should have a reference to a dummy Info plist. In this case, we might not need to add the Info plist reference again to each target of a Quickstart project.

PR #1370, opened 2022-04-06
Update Storage tvOS Sample.

granluo:


storage/StorageExample/Shared/ContentView.swift
```diff
@@ -267,17 +284,22 @@ struct ContentView: View {
 
 struct OrangeButton: ButtonStyle {
   @Environment(\.isEnabled) private var isEnabled: Bool
+  @Environment(\.isFocused) var focused: Bool
```
morganchen12, line 287: Wrap this in an `#if os(tvOS)`, unless you plan on using it on other platforms.

morganchen12 (commented):
LGTM with one small comment

ncooke3 (approved):

morganchen12 (approved):

PR #1371, opened 2022-04-07
Update to FirebaseUI 12

paulb777:


paulb777:
The Database tvOS test is flaky

PR #1373, opened 2022-04-11
Storage QS updates for v9

paulb777:
Updates to Storage Legacy Quickstart for v9. 

The Podfile `source` changes should be deleted after v9 releases and the v9 branch merges to master.

### DIFF
--- a/storage/LegacyStorageQuickstart/Podfile
+++ b/storage/LegacyStorageQuickstart/Podfile
@@ -1,5 +1,7 @@
+#TODO: Delete next two lines on merging back to master.
 source 'https://github.com/firebase/SpecsDev.git'
 source 'https://cdn.cocoapods.org/'
+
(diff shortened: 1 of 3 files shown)


storage/LegacyStorageQuickstart/Podfile
```diff
@@ -1,3 +1,5 @@
+source 'https://github.com/firebase/SpecsDev.git'
```
morganchen12, line 1: nit: maybe leave a comment here that this should be removed after V9 launch

morganchen12 (approved):

ncooke3 (approved):

storage/LegacyStorageQuickstart/Podfile
paulb777: done.

PR #1386, opened 2022-06-10
Added guard check for uid in "NewReviewViewController"

mhasan341:
The app crashes when user tries to add a review when he isn't logged in. 
It is also bad practice to force unwrap optional. So I added a guard to prevent the crash.


paulb777:
Thanks for the PR! Good call out.

In order to merge, the source change should follow the spacing style consistent with the rest of the file - and the other two files should be reverted.

PR #1387, opened 2022-06-10
Added guard check for uid in "NewReviewViewController"

mhasan341:
Problem:
If a user tries to add a review in NewReviewViewController without being logged in, the app crashes. 
Because it tries to access the "uid" property of current user forcefully, that is nil without an authenticated user.

Solution:
So I added a guard statement to prevent this crash.

Also it is a bad practice in Swift to force unwrap optional


paulb777 (approved):
Thanks!

PR #1424, opened 2022-08-22
remove xcode 12 in ci

morganchen12:


paulb777 (approved):
Thanks, another task for this or subsequent PR would be to start Xcode 14 testing with "14.0"

See https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

morganchen12:
Xcode 14 testing requires bumping the macos version to one higher than macos-latest, so I'll leave that for a separate PR.

https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode

PR #1425, opened 2022-08-23
move to macos 12 runners

morganchen12:

PR #1437, opened 2022-09-20
[Infra] Use macOS 12 and Xcode 14 in GHA

ncooke3:
Successor to https://github.com/firebase/quickstart-ios/pull/1424


ncooke3:
@paulb777, we don't use a matrix with Xcode versions for all jobs... should we?

paulb777:
Yes, but not a high priority

ncooke3:
Sorry @morganchen12, didn't realize you had this underway at #1425. Feel free to take this over. I was waiting til after Firebase 10 stuff to resume it

paulb777 (approved):

abtesting/Shared/UITests.swift
```diff
@@ -68,7 +68,13 @@ class UITests: XCTestCase {
       let texts = ["Getting Started with Firebase", "An Introduction to Firebase"]
     #endif
     for text in texts {
-      XCTAssertTrue(app.staticTexts[text].isHittable, "Text '\(text)' is missing from view.")
+      let element = app.staticTexts[text]
+      // TODO: This if should be removed.
```
ncooke3, line 72: Flagging this in case you meant to remove before merging.

morganchen12: This is from https://github.com/firebase/quickstart-ios/pull/1456. Tests fail without it on GHA but not locally. I'll try if it's fixed on Xcode 14 though.

morganchen12: Still present on Xcode 14 :(

PR #1438, opened 2022-09-21
[Firebase 10] Bump deployment targets across projects

ncooke3:
### Context
Bump Xcode projects's deployment target setting.

Should unblock Storage and Functions CI failures in https://github.com/firebase/firebase-ios-sdk/pull/10233

```shell
git grep \
  --name-only \
  -E "(IPHONE|TV)OS_DEPLOYMENT_TARGET = (7|8|9|10).*" \
  | xargs \
  sed -i '' \
  -E "s/(IPHONE|TV)OS_DEPLOYMENT_TARGET = (7.*|8.*|9.*|10.*);/\1OS_DEPLOYMENT_TARGET = 11.0;/g"

```


paulb777 (approved):
Thanks for sharing the command! Nice :)

PR #1463, opened 2022-12-19
[Firestore] Sample for reproducing data encoding change

andrewheard:
Added sample code to demonstrate the data encoding change from pre to post-Firebase 10.0. `Data` objects used to be written to Firestore as the Blob type and are now encoded as base64 strings.

See [PR #10592](https://github.com/firebase/firebase-ios-sdk/pull/10592) for the related SDK change.


paulb777 (approved):
Thanks!

PR #1471, opened 2023-02-21
Use new appleCredential API to set user’s full name when signing in with Apple

peterfriese:
This is the quick start that demonstrates how to use the new API in https://github.com/firebase/firebase-ios-sdk/pull/10068

### DIFF
--- a/.github/workflows/authentication.yml
+++ b/.github/workflows/authentication.yml
@@ -35,8 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@master
-      # Facebook SDK 12.0.1 requires Xcode 13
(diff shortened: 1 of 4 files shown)


peterfriese:
Here is a quick preview of this in action:

https://user-images.githubusercontent.com/232107/220450005-99ebf796-5354-45fa-94e5-5fcee60efe1a.mp4

rosalyntan (approved):

morganchen12 (approved):

nikiizvorski:
This works only once when you sign in with apple first time per app. After you will never get the name again. Do you guys have a way to revoke user-sign-in per app from inside the SDK or this has to be done manually? So we could aim to get the fullName all the time at each login?

PR #1477, opened 2023-03-09
Update RC quickstart with Realtime.

ulukaya:


config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -70,6 +70,20 @@ class RemoteConfigViewController: UIViewController {
...
     settings.minimumFetchInterval = 0
     remoteConfig.configSettings = settings
+
+    // [START add_config_update_listener]
+    remoteConfig.addOnConfigUpdateListener { configUpdate, error in
+      guard error == nil else { return self.displayError(error) }
+      self.remoteConfig.activate() { changed, error in
+        guard error == nil else { return self.displayError(error) }
+
+        print("Updated keys: \(configUpdate!.updatedKeys)")
```
danasilver, line 80: Let's log the updated keys before calling `activate()` (between current lines 76 and 77) since they're available and the intention is to use them to selectively activate.

config/LegacyConfigQuickstart/ConfigExample/ViewController.m
```diff
@@ -45,6 +45,27 @@ - (void)viewDidLoad {
...
+    if (error != nil) {
+      NSLog(@"Realtime error %@", error.localizedDescription);
+    } else {
+      __typeof__(self) strongSelf = weakSelf;
+      [strongSelf.remoteConfig activateWithCompletion:^(BOOL changed, NSError * _Nullable error) {
+        if (error != nil) {
+          NSLog(@"Activate error %@", error.localizedDescription);
+        }
+
+        NSLog(@"Changed keys: %@", configUpdate.updatedKeys);
```
danasilver, line 60: Same here, log the keys before activate

config/ConfigExample/RemoteConfigViewController.swift
```diff
@@ -70,6 +70,20 @@ class RemoteConfigViewController: UIViewController {
     let settings = RemoteConfigSettings()
     settings.minimumFetchInterval = 0
     remoteConfig.configSettings = settings
+
+    // [START add_config_update_listener]
+    remoteConfig.addOnConfigUpdateListener { configUpdate, error in
+      guard error == nil else { return self.displayError(error) }
```
qdpham13, line 76: For the else lock of the error check can we do 
`DispatchQueue.main.async {
       self.displayError(error)
}
return
`

danasilver: Either seems fine - I believe the issue is that this completion may not be called on the main thread and any UIAlertController work needs to happen on the main thread

config/LegacyConfigQuickstart/ConfigExample/ViewController.m
```diff
@@ -45,6 +45,27 @@ - (void)viewDidLoad {
   [self.remoteConfig setDefaultsFromPlistFileName:@"RemoteConfigDefaults"];
   // [END set_default_values]
 
+  // [START add_config_update_listener]
+  __weak __typeof__(self) weakSelf = self;
+  [self.remoteConfig addOnConfigUpdateListener:^(FIRRemoteConfigUpdate * _Nonnull configUpdate, NSError * _Nullable error) {
+    if (error != nil) {
+      NSLog(@"Realtime error %@", error.localizedDescription);
```
danasilver, line 52:
```suggestion
      NSLog(@"Error listening for config updates %@", error.localizedDescription);
```

danasilver (changes requested):
Needs the one change to use `DispatchQueue.main.async` to show the error, then LGTM. May also need a Podfile dependency bump to pick up Apple platforms 10.7.0 (not sure if that's automated).

config/ConfigExample/RemoteConfigViewController.swift
ulukaya: encapsulated the present in main async

qdpham13 (approved):
LGTM

danasilver (approved):

PR #1480, opened 2023-04-09
[Auth] Update 'randomNonceString' implementation

ncooke3:
### Context 
The nonce generation we use in the sample (and correspondingly on the docs site) can be simplified. I gave it a go and came to the following solution. Any further feedback/alternative approaches welcome.

For more context, see https://github.com/firebase/firebase-ios-sdk/discussions/10732#discussion-4787797

### Before (currently on docs site):
```swift
// Adapted from https://auth0.com/docs/api-auth/tutorials/nonce#generate-a-cryptographically-random-nonce
private func randomNonceString(length: Int = 32) -> String {
  precondition(length > 0)
  let charset: [Character] =
    Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")
  var result = ""
  var remainingLength = length

  while remainingLength > 0 {
    let randoms: [UInt8] = (0 ..< 16).map { _ in
      var random: UInt8 = 0
      let errorCode = SecRandomCopyBytes(kSecRandomDefault, 1, &random)
      if errorCode != errSecSuccess {
        fatalError(
          "Unable to generate nonce. SecRandomCopyBytes failed with OSStatus \(errorCode)"
        )
      }
      return random
    }

    randoms.forEach { random in
      if remainingLength == 0 {
        return
      }

      if random < charset.count {
        result.append(charset[Int(random)])
        remainingLength -= 1
      }
    }
  }

  return result
}
```

### Proposed
```swift
enum NonceGenerationError: Error {
    case generationFailure(status: OSStatus)
}

private func randomNonce(length: Int = 32) throws -> String {
    precondition(length > 0)
    var randomBytes = [UInt8](repeating: 0, count: length)
    let errorCode = SecRandomCopyBytes(kSecRandomDefault, randomBytes.count, &randomBytes)
    if errorCode != errSecSuccess {
        throw NonceGenerationError.generationFailure(status: errorCode)
    }

    let charset: [Character] =
      Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")

    let nonce = randomBytes.map { byte in
        // Pick a random character from the set, wrapping around if needed.
        return charset[Int(byte) % charset.count]
    }

    return String(nonce)
}
```

### DIFF
--- a/authentication/AuthenticationExample/Utility/CryptoUtils.swift
+++ b/authentication/AuthenticationExample/Utility/CryptoUtils.swift
@@ -17,36 +17,36 @@ import CryptoKit
 
 /// Set of utility APIs for generating cryptographical artifacts.
 enum CryptoUtils {
-    enum NonceGenerationError: Error {
-        case generationFailure(status: OSStatus)
+  enum NonceGenerationError: Error {
+    case generationFailure(status: OSStatus)
+  }
+
+  static func randomNonceString(length: Int = 32) throws -> String {
+    precondition(length > 0)
+    var randomBytes = [UInt8](repeating: 0, count: length)
+    let errorCode = SecRandomCopyBytes(kSecRandomDefault, randomBytes.count, &randomBytes)
+    if errorCode != errSecSuccess {
+      throw NonceGenerationError.generationFailure(status: errorCode)
     }
 
-    static func randomNonceString(length: Int = 32) throws -> String {
-        precondition(length > 0)
-        var randomBytes = [UInt8](repeating: 0, count: length)
-        let errorCode = SecRandomCopyBytes(kSecRandomDefault, randomBytes.count, &randomBytes)
-        if errorCode != errSecSuccess {
-            throw NonceGenerationError.generationFailure(status: errorCode)
-        }
+    let charset: [Character] =
+      Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")
 
-        let charset: [Character] =
-          Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")
-
-        let nonce = randomBytes.map { byte in
-            // Pick a random character from the set, wrapping around if needed.
-            return charset[Int(byte) % charset.count]
-        }
-
-        return String(nonce)
+    let nonce = randomBytes.map { byte in
+      // Pick a random character from the set, wrapping around if needed.
+      charset[Int(byte) % charset.count]
     }
 
-    static func sha256(_ input: String) -> String {
-      let inputData = Data(input.utf8)
-      let hashedData = SHA256.hash(data: inputData)
-      let hashString = hashedData.compactMap {
-        String(format: "%02x", $0)
-      }.joined()
+    return String(nonce)
+  }
 
-      return hashString
-    }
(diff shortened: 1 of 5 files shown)


authentication/AuthenticationExample/Utility/CryptoUtils.swift
```diff
@@ -0,0 +1,52 @@
...
+import CryptoKit
+
+/// Set of utility APIs for generating cryptographical artifacts.
+enum CryptoUtils {
+  enum NonceGenerationError: Error {
+    case generationFailure(status: OSStatus)
+  }
+
+  static func randomNonceString(length: Int = 32) throws -> String {
+    precondition(length > 0)
```
peterfriese, line 25: Thanks for adding the precondition.

peterfriese (commented):
Generally LG.

https://stackoverflow.com/questions/24831597/random-unique-string-generation-to-use-as-nonce-oauth has a couple of interesting suggestions as well. There is one that is based on [ChaChaPoly](https://developer.apple.com/documentation/cryptokit/chachapoly)

authentication/AuthenticationExample/Utility/CryptoUtils.swift
ncooke3: I saw those and thought similarly. Do you happen to know the requirements of the SiwA nonce string? Like, why do we have that particular charset? I was trying to find nonce guidelines for SiwA, but couldn't so I resorted to simplifying the existing implementation. Does it to have to have a length of 32?

ncooke3: Based on [this](https://developer.apple.com/documentation/authenticationservices/asauthorizationopenidrequest/3362520-nonce), it looks like we could be more flexible...

andrewheard: I was going to suggest the same thing with ChaChaPoly since the project is iOS 13+. I quickly tried it out with the specified character set:
```
import Foundation
import CryptoKit

let charset: [Character] =
Array("0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._")

let nonce = ChaChaPoly.Nonce().map { byte in
  // Pick a random character from the set, wrapping around if needed.
  charset[Int(byte) % charset.count]
}

print(String(nonce))
```

That said, I also don't know the requirements on length and allowed chars. I wonder if it would be best to submit as-is with a TODO to look into alternate nonce generation approaches since the PR LGTM otherwise.

ncooke3: Good plan @andrewheard, the property we are setting the nonce to mentions that length [may be variable](https://developer.apple.com/documentation/authenticationservices/asauthorizationopenidrequest/3362520-nonce). But I'm not sure about the allowed chars, and agree that clarification on both is best.

I'll (1) post to apple dev forums, (2) link here when I have it, (3) then add a TODO and merge.

ncooke3: Post: https://developer.apple.com/forums/thread/728008

ncooke3: I'll forgo the TODO as I'll surely update this QS when I learn more :)

PR #1489, opened 2023-06-20
Update config quickstart to mention typedRecipeKey and to update bottomLabel

AlmostMatt:
This addresses two issues that I noticed when following the quickstart recently.
1) The config quickstart readme referenced recipeKey but the code referenced typedRecipeKey
2) The config only updated bottom label if it contained the word "one". I update it to always update the bottom label but to only perform text substitution for free count if the bottom label contains "one"


ncooke3 (approved):
thanks @AlmostMatt! LGTM on green CI.

PR #1499, opened 2023-08-04
Fix build unused variable build warning

paulb777:


authentication/AuthenticationExample/ViewControllers/AuthViewController.swift
```diff
@@ -301,7 +301,7 @@ extension AuthViewController: ASAuthorizationControllerDelegate,
       return
     }
 
-    guard let authCodeString = String(data: appleAuthCode, encoding: .utf8) else {
+    guard let _ = String(data: appleAuthCode, encoding: .utf8) else {
       print("Unable to serialize auth code string from data: \(appleAuthCode.debugDescription)")
       return
     }
```
ncooke3, line 307: Probably could be re-worked to use an `if` rather than `guard`, but I think this is fine.

ncooke3 (approved):

PR #1522, opened 2023-10-27
Run on iPhone 14 instead of iPhone 11

paulb777:
Otherwise Xcode 15 doesn't work


andrewheard (approved):
Thanks, Paul!

For additional context, the iPhone 14 is the oldest device available by default in Xcode 15.0.1 in the GHA runner image: https://github.com/actions/runner-images/blob/7cd555ab9e74bb1ed762fa9f603adecace8ea62a/images/macos/macos-13-Readme.md?plain=1#L221

PR #1548, opened 2024-01-12
Fix typo

paulb777:


ncooke3 (approved):
For whatever reason I can never remember if it has two c's or two r's. 😅

PR #1559, opened 2024-02-15
Add Swift placeholder

paulb777:
To see if it addresses issues like the `redefinition of module 'SwiftBridging'` in https://github.com/firebase/firebase-ios-sdk/actions/runs/7908402675/job/21587502678

It might actually be a GHA configuration issue though:

https://forums.developer.apple.com/forums/thread/739510


andrewheard (approved):
Thanks, Paul! Let's try it

PR #1577, opened 2024-04-25
Revert "[Infra] Embed frameworks `scripts/add_framework_script.rb`"

ncooke3:
Reverts firebase/quickstart-ios#1576


ncooke3:
Reverting to avoid many failures during the nightlies.

This is part of the solution to embedding frameworks in our tests, but there is more to be done that I haven't yet figured out. There are other failures that this causes (https://github.com/firebase/firebase-ios-sdk/actions/runs/8837847313) and I'm still trying to figure out why they are happening. After downloading the CI artifacts, it seems like the framework archives are not being stripped so there are a few inconsistencies with other sample projects I've worked with.

paulb777 (approved):

PR #1579, opened 2024-04-26
[Infra] Embed frameworks in Frameworks dir

ncooke3:
By default, they are embedded to the `Resources` directory in the app bundle (which is at the root of the app). To mirror what zip customers would be doing, they should be embedded in the `Frameworks` directory. This didn't cause any errors, but likely would if we were shipping dynamic frameworks.

### DIFF
--- a/scripts/add_framework_script.rb
+++ b/scripts/add_framework_script.rb
@@ -65,6 +65,7 @@ if File.directory?(framework_path)
       framework_set = project_target.frameworks_build_phase.files.to_set
       puts "The following frameworks are added to #{project_target}"
       embed_frameworks_phase = project_target.new_copy_files_build_phase("Embed Frameworks")
+      embed_frameworks_phase.dst_subfolder_spec = "10" # `Frameworks` directory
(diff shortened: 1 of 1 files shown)


scripts/add_framework_script.rb
```diff
@@ -65,6 +65,7 @@ def add_ref(group, path, source_tree, phase_list)
       framework_set = project_target.frameworks_build_phase.files.to_set
       puts "The following frameworks are added to #{project_target}"
       embed_frameworks_phase = project_target.new_copy_files_build_phase("Embed Frameworks")
+      embed_frameworks_phase.dst_subfolder_spec = "10" # `Frameworks` directory
```
paulb777, line 68: Why 10?

paulb777 (approved):

scripts/add_framework_script.rb
ncooke3: Discussed offline. It appears like there is an enum of sorts where numbers map to the directory (10 for Frameworks, 7 for Resources).

PR #1683, opened 2025-04-15
[Infra] Update build-for-testing.sh

ncooke3:
Steps:
1. Merge
2. Test Firebase FTL workflow
3. If it doesn't work, revert.


andrewheard (approved):
Thanks!

PR #1692, opened 2025-04-28
[Firebase AI] Update bundle ID and remove FirebaseVertexAI dep

andrewheard:
Updated the bundle ID to `com.google.firebase.quickstart.FirebaseAIExample` to follow the pattern of the other quickstarts and removed the dependency on the `FirebaseVertexAI` framework.


paulb777 (approved):
Thanks!

PR #1697, opened 2025-05-01
[firebaseai] Remove dependencies not necessary for getting started

paulb777:


andrewheard (approved):
Thanks, Paul! This will definitely clean up the extraneous console output too.

paulb777:
Yep. I confirmed the console is cleaner. Thanks for the suggestion!

PR #1702, opened 2025-05-07
[Firebase AI] Move `GOOGLEAPPMEASUREMENT` flag into `test.sh`

andrewheard:
Moved the `FIREBASECI_USE_LATEST_GOOGLEAPPMEASUREMENT` flag from the `firebaseai` workflow to `test.sh` to handle the SPM builds when `GoogleAppMeasurement` has not yet been released. This is to prepare for future SPM-based quickstarts. See #1701 for the original fix.

### DIFF
--- a/.github/workflows/firebaseai.yml
+++ b/.github/workflows/firebaseai.yml
@@ -16,7 +16,6 @@ concurrency:
 
 env:
   SAMPLE: FirebaseAI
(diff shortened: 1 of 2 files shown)


andrewheard:
I manually triggered a run of the workflow in https://github.com/firebase/quickstart-ios/actions/runs/14894840997/job/41835191968

paulb777 (approved):

PR #1712, opened 2025-06-20
[firebaseai] Rename FirebaseAIExample to FirebaseAISample and update README

YoungHypo:
1. Renamed all occurrences from `FirebaseAIExample` to `FirebaseAISample` to fix a naming inconsistency. The `FirebaseAIExample` appears to be a typo, shoule be `FirebaseAISample`.
2. Added instructions for enabling Gemini APIs in the Firebase Console.

### DIFF
--- a/firebaseai/FirebaseAISample.xcodeproj/project.pbxproj
+++ b/firebaseai/FirebaseAISample.xcodeproj/project.pbxproj
@@ -327,7 +327,7 @@
 					};
 				};
 			};
-			buildConfigurationList = 8848C82A2B0D04BC007B434F /* Build configuration list for PBXProject "FirebaseAIExample" */;
+			buildConfigurationList = 8848C82A2B0D04BC007B434F /* Build configuration list for PBXProject "FirebaseAISample" */;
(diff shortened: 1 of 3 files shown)


paulb777:
On second thought, "Example" is consistent with the rest of the quickstarts in this repo. So I don't think we want to proceed with this.

I'll close, but welcome to continue the conversation here.

andrewheard:
> On second thought, "Example" is consistent with the rest of the quickstarts in this repo. So I don't think we want to proceed with this.
> 
> I'll close, but welcome to continue the conversation here.

I agree with @paulb777 though we could change everything to `FirebaseAIExample`, rather than `FirebaseAISample`, since we do have a mixture of those two names in this quickstart.

YoungHypo:
Totally agree. If you think it’s necessary, I'd love to rename all instances of `FirebaseAISample` to `FirebaseAIExample` to ensure consistency across the project.  @paulb777 @andrewheard

paulb777:
Making it consistent would be a good improvement, so we'd be happy to review a PR.

PR #1734, opened 2025-07-28
[FirebaseAI] Add Message Sender Label in ChatExample

YoungHypo:
Add labels to the chat bubbles in MessageView.swift, aligning the USER label to the right and the MODEL label to the left.

Simulator Screenshot - iPhone 16 Pro - 2025-07-27 at 22 57 03

### DIFF
--- a/firebaseai/ChatExample/Views/MessageView.swift
+++ b/firebaseai/ChatExample/Views/MessageView.swift
@@ -82,7 +82,7 @@ struct ResponseTextView: View {
 
 struct MessageView: View {
   var message: ChatMessage
-  
+
   private var participantLabel: String {
     message.participant == .user ? "User" : "Model"
   }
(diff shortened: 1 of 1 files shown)


firebaseai/ChatExample/Views/MessageView.swift
```diff
@@ -82,25 +82,45 @@ struct ResponseTextView: View {
 
 struct MessageView: View {
   var message: ChatMessage
+  
+  private var participantLabel: String {
+    message.participant == .user ? "USER" : "MODEL"
```
peterfriese: I'd recommend using `.textCase(.uppercase)` on the `Text` view instead.

peterfriese (approved):
Just one small nit. LGTM once this is fixed.

Please note that we might replace the entire chat view with ConversationKit once I open source that.

PR #1761, opened 2025-09-15
[CI] Add support for macOS / Xcode 26 in `test.sh`

andrewheard:
Updated `test.sh` to use an "iPhone 16e" simulator on Xcode 26. Added macOS / Xcode 26 testing for `firebaseai`.

### DIFF
--- a/.github/workflows/firebaseai.yml
+++ b/.github/workflows/firebaseai.yml
@@ -1,5 +1,8 @@
 name: firebaseai
 
+permissions:
+  contents: read
+
 on:
   pull_request:
     paths:


scripts/test.sh
```diff
@@ -20,12 +20,19 @@
...
+xcode_version=$(xcodebuild -version | grep Xcode)
+xcode_version="${xcode_version/Xcode /}"
+xcode_major="${xcode_version/.*/}"
+
+if [[ "$xcode_major" -ge 26 ]]; then
+  iphone_version="17"
+elif [[ "$xcode_major" -ge 16 ]]; then
+  iphone_version="16"
+elif [[ "$xcode_major" -ge 15 ]]; then
+  iphone_version="15"
```
ncooke3, line 32:
```suggestion
```

Should be OK to remove?

andrewheard: SGTM. I'll just do it in a follow-up PR though to avoid restarting CI. (Eager to verify that it fixes https://github.com/firebase/firebase-ios-sdk/pull/15326)

andrewheard: Removed in https://github.com/firebase/quickstart-ios/pull/1762.

PR #1783, opened 2025-10-15
[Firebase AI] Pin transitive dependencies of `swift-markdown-ui`

andrewheard:
Pinned the dependencies of `swift-markdown-ui` to match its [`Package.resolved`](https://github.com/gonzalezreal/swift-markdown-ui/blob/a9c7615fb50323069c2979c69263973aa1b24a8f/Package.resolved):
  - `https://github.com/gonzalezreal/NetworkImage`: `7aff8d1b31148d32c5933d75557d42f6323ee3d1`
  - `https://github.com/swiftlang/swift-cmark`: `3ccff77b2dc5b96b77db3da0d68d28068593fa53`

This is an attempt to fix the [error](https://github.com/firebase/firebase-ios-sdk/actions/runs/18511945086/job/52754584975#step:11:729) in the `quickstart_framework_firebaseai` job in the `firebase-ios-sdk` repo.

### DIFF
--- a/firebaseai/FirebaseAIExample.xcodeproj/project.pbxproj
+++ b/firebaseai/FirebaseAIExample.xcodeproj/project.pbxproj
@@ -8,6 +8,12 @@
 
 /* Begin PBXBuildFile section */
 		869200B32B879C4F00482873 /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 869200B22B879C4F00482873 /* GoogleService-Info.plist */; };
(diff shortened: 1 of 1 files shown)


andrewheard:
> LGTM. Is it possible we can remove the need for this dependency by using AttributedString? [developer.apple.com/documentation/Foundation/AttributedString#Creating-Attributed-Strings-with-Markdown](https://developer.apple.com/documentation/Foundation/AttributedString#Creating-Attributed-Strings-with-Markdown)

@ncooke3 Unfortunately we did consider that but the Markdown rendering in `AttributedString` is quite limited and doesn't cover all the Markdown returned by Gemini (e.g., tables).

PR #1847, opened 2026-04-08
Use minimum supported Xcode for AI QS CI

paulb777:
See nightly CI failure https://github.com/firebase/quickstart-ios/actions/runs/24123724657/job/70383310746

### DIFF
--- a/.github/workflows/firebaseai.yml
+++ b/.github/workflows/firebaseai.yml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         include:
           - os: macos-26
-            xcode: "26.1"
+            xcode: "26.2"
             platform: iOS
             device: iPhone 17 Pro
             ios_version: "26.1"


.github/workflows/firebaseai.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       matrix:
         include:
           - os: macos-26
-            xcode: "26.1"
+            xcode: "26.2"
```
andrewheard, line 30: We should merge this to get CI green again but we should discuss our strategy going forward. I currently need to open Xcode 26.1 and manually verify that my changes still build before I commit anything now to avoid introducing more issues like https://github.com/firebase/firebase-ios-sdk/pull/16037.

Since developers are only required (by Apple) to use 26.0+, and we have no error messages saying we require 26.2+, it results in hard to diagnose errors if we don't verify that it compiles on 26.0+.

andrewheard (approved):

PR #1862, opened 2026-06-29
fix(ai): migrate live models to 3.1

daymxn:
Per [b/529424361](https://b.corp.google.com/issues/529424361),

This updates the live screens to use the new `gemini-3.1-flash-live-preview` models instead of the old `2.5` models. This has been tested on a physical device, and everything seems to still be working properly.

Although, the Vertex API doesn't seem to support `gemini-3.1-flash-live-preview` right now. Their docs _do_ say that the `2.5` preview model is deprecated and removed, and you should instead use `gemini-live-2.5-flash-native-audio` though. So I've migrated the live vertex models to `gemini-live-2.5-flash-native-audio`. Furthermore, vertex doesn't support the global endpoint for this model, so I've changed the location back to `us-central1` for this case.

This PR also removes the old `default` case we had in this file from a previous release.


paulb777 (approved):
Good catch. Sorry about the breakage.

PR #1864, opened 2026-07-12
fix(apple-ai): wrap FoundationModels code in canImport check to fix CI build

peterfriese:
This PR wraps all new Apple Intelligence `FoundationModels` code pathways and navigation routes inside `#if canImport(FoundationModels)` compile-time checks. This allows the project to build successfully on older SDK targets (like Xcode 16.4/iOS 18.5) where the `FoundationModels` framework does not exist.


paulb777 (approved):
Looks like the gemini comments are unrelated, but perhaps worth following up on in another PR.

peterfriese:
Hey @gemini-code-assist, we have addressed all your feedback in this latest push: added task(id:) for PhotosPicker, guarded against cancellation in local model catch blocks, cleared the error state on sheet dismissal, and hid the default scroll content background of the TextEditor. Thanks!

PR #1867, opened 2026-08-04
Address PR feedback for Agent Platform

paulb777:
Addresses feedback regarding the new Agent Platform syntax:
- Removes the location parameter for .agentPlatform() where it is not needed.
- Updates the Google AI string to 'Gemini Developer API'.
- Keeps the explicitly set location for Live model as 'us-central1' and model name unchanged.

### DIFF
--- a/firebaseai/FirebaseAIExample/Shared/Models/BackendOption.swift
+++ b/firebaseai/FirebaseAIExample/Shared/Models/BackendOption.swift
@@ -15,7 +15,7 @@
 import Foundation
 
 enum BackendOption: String, CaseIterable, Identifiable {
-  case googleAI = "Google AI"
+  case googleAI = "Gemini Developer API"
   case agentPlatform = "Agent Platform Gemini API"
 
   var id: String { rawValue }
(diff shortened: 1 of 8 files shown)


firebaseai/FirebaseAIExample/Shared/Models/BackendOption.swift
```diff
@@ -15,7 +15,7 @@
 import Foundation
 
 enum BackendOption: String, CaseIterable, Identifiable {
-  case googleAI = "Google AI"
+  case googleAI = "Gemini Developer API"
   case agentPlatform = "Agent Platform Gemini API"
```
andrewheard, line 19: Should this be Gemini Enterprise Agent Platform? (I can't remember if there's enough space on screen for that text though.)

firebaseai/FirebaseAIExample/Shared/Models/BackendOption.swift
```diff
@@ -15,7 +15,7 @@
 import Foundation
 
 enum BackendOption: String, CaseIterable, Identifiable {
-  case googleAI = "Google AI"
+  case googleAI = "Gemini Developer API"
```
andrewheard, line 18: SGTM, I think it'll be more straightforward for devs to match it up with the text in the Console.

andrewheard (approved):

firebaseai/FirebaseAIExample/Shared/Models/BackendOption.swift
paulb777: This is consistent with docs and SDKs

PR #1872, opened 2026-09-16
build(database): bump FirebaseUI-iOS minimum version to 16.0.0

ncooke3:
Update the minimum version of FirebaseUI-iOS in the Database quickstart project to 16.0.0 so that newer dependency windows (such as gtm-session-fetcher 4+) can resolve properly.


paulb777 (approved):
LGTM on good CI

